PR #338, opened 2017-07-25
Update appveyor.yml

aymanhab:
Try to update modules before building


chrisdembia:
@aymanhab this is ready to be reviewed.

This PR now uploads binaries to AppVeyor, which are available [here](https://ci.appveyor.com/project/opensim-org/opensim-gui/build/1.0.741/artifacts):

screen shot 2017-10-05 at 7 30 48 pm

I have been able to download these binaries and run them on Windows and they seem to work.

TODO:
- Doxygen is not included in the binary, because opensim-core's AppVeyor does not yet build doxygen.
- Models are not in the resulting distribution, because the repo is private.
- If desired, I can also upload these binaries to a publicly available website (sourceforge). That'll make it even easier to share binaries with potential testers. (The AppVeyor download requires you to be a part of our AppVeyor account).

chrisdembia:
The Travis builds won’t pass until opensim-core ‘s Mac build includes java bindings. I’ll be uploading the Mac GUI build to source forge ( because Travis does not offer hosting)

chrisdembia:
> We should figure out how to make models public so that the distro is self contained and useful.

I think we can handle this in a separate PR. I've pinged multiple people about this over in https://github.com/opensim-org/opensim-models/issues/1

chrisdembia:
Thanks @aymanhab

PR #409, opened 2017-10-09
Build Mac GUI on Travis

chrisdembia:
Travis now pushes a Mac installer for the GUI to https://sourceforge.net/projects/myosin/files/opensim-gui/

I was able to download this PKG and install it on my Mac.

This PR is *not* ready to be merged.

TODO:
- [x] Get https://github.com/opensim-org/opensim-core/pull/1964 merged.
- [x] Get https://github.com/opensim-org/opensim-models/issues/1 resolved.
- [x] The GUI installer created with Travis does not currently set the name of the userdir correctly.


chrisdembia:
I've finished all 3 TODO items above and this is now ready for review.

aymanhab:
@chrisdembia Instead of using post install to change userdir, would it be possible to use ANT and modify the file in the etc folder directly so that it can be used cross platform.

One other question, not clear where do you get opensim-core, and if not from the submodule, shouldn't we just remove it? Thanks

chrisdembia:
> @chrisdembia Instead of using post install to change userdir, would it be possible to use ANT and modify the file in the etc folder directly so that it can be used cross platform.

Not quite. The postinstall script is run on the user's machine during the installation process, and is only necessary because of bugs in the GUI (without that postinstall script, the GUI will hang on first load). I would like to get rid of the postinstall script.

You're right though that we could use ANT to edit the version string in all of the Mac installation files. I'll do that.

> One other question, not clear where do you get opensim-core, and if not from the submodule, shouldn't we just remove it? 

It's not from the submodule, it's from sourceforge.net/projects/myosin. BInaries for opensim-core are uploaded to sourceforge as part of opensim-core's travis build. I'll remove that submodule.

chrisdembia:
I edited build.xml to make the PKG, but the resulting PKG doesn't work correctly (the installer says the installation fails). I'll have to look into this.

chrisdembia:
@aymanhab I cleaned up how the PKG is made (it's now done by ANT). This is now ready for review. 

The latest PKG is available here:   -DCMAKE_PREFIX_PATH=/Users/chris/nmblprojects/tomu/muscollo_dependencies_install/opensim-core;/Users/chris/nmblprojects/tomu/muscollo_dependencies_install/colpack -DCMAKE_INSTALL_PREFIX=/Users/chris/nmblprojects/tomu/muscollo-install

aymanhab:
Looks great @chrisdembia Thank you :+1:

chrisdembia:
Thanks @aymanhab

Issue #404, opened 2017-10-06
Reference jxbrowser in readme


aymanhab:
Per communication with jxbrowser team, add note to repository.

chrisdembia:
Yes I still plan on doing this. Thanks for the reminder

aymanhab:
I believe this can be closed now @chrisdembia

PR #412, opened 2017-10-10
First pass at complying with third party licenses.

chrisdembia:
@aymanhab @jenhicks 

This is my first attempt at complying with various licenses of third party software used by the GUI. I also tried to follow Apache's guidelines [here](http://www.apache.org/dev/licensing-howto.html) for the content of the LICENSE and NOTICE file.

- Added/edited the LICENSE and NOTICE files in the repo
- Created LICENSE and NOTICE files that get installed into the binary distribution
  - The reason for this additional pair of files is explained at the link above (some dependencies only appear in the binary distribution).
- Edited the LICENSE text that appears in the Mac installer.
- Deleted the old Copyright.txt file.
- Cleaned up the README (removed the bit about GUI users being forced to cite a paper).
- Carried over the JxBrowser notice (#404)
- Edited the About/Acknowledgements panel in the GUI (see the third pane):

screen shot 2017-10-09 at 11 02 06 pm

There is more to be done, namely complying with the "dependencies of dependencies" (OpenSim's and Simbody's dependencies) and copying over the content of opensim-core's NOTICE. We can handle these later.

EDIT: 

- I've also updated Confluence's Acknowledgements page: https://simtk-confluence.stanford.edu/display/OpenSim/Acknowledgements
- [This webpage](https://www.teamdev.com/jxbrowser/lgpl-compliance) provides a handy list of what one should do to comply with the LGPL.


jenhicks:
Thanks @chrisdembia. These changes look good to me. Could you add the text requesting acknowledgement to the GUI acknowledgements tab, as well?

chrisdembia:
@jenhicks I've attempted to address your feedback. Here's what the acknowledgements tab looks like now:

screen shot 2017-10-10 at 10 40 46 am

I edited most of the text so that it fits in the window.

Here's what the window looks like without this PR:

screen shot 2017-10-10 at 10 44 39 am

jenhicks:
Looks good. Thanks for shortening up the text ... that makes sense to do.

chrisdembia:
It would be good to merge this before we get merge conflicts or the code base diverges too much.

aymanhab:
Thanks much @chrisdembia, a few comments:

- Although the code still contains vtk we're not "using it" so I think we'd be ok removing that. 
- I also don't know about JxBrowser and three.js which we actually use even though not mentioned on this dialog. 
Any thoughts/suggestions?

chrisdembia:
@aymanhab you're specifically talking about the About dialog, correct? If so, then I agree with your suggestions.

aymanhab:
Yes, I mean the About dialog not the readme or any github issues since by official release we should have the entanglements removed, I hope 😃

chrisdembia:
@aymanhab I've updated the 3rd text box on the Acknolwedgments tab:

screen shot 2017-11-01 at 10 54 37 pm

Ready for review again.

chrisdembia:
Thanks @aymanhab

Issue #416, opened 2017-10-13
[AppVeyor] Automatically set userdir to use commit hash and date.


chrisdembia:

jenhicks:
Closing since @aymanhab believes this is done .. please feel free to re-open, if not

chrisdembia:
This is done for Travis, but not yet for AppVeyor. See [here](https://github.com/opensim-org/opensim-gui/blob/master/appveyor.yml#L57).

aymanhab:
Don't know if this works for windows

jenhicks:
@aymanhab thinks this is fixed since we now automate the build. Please reopen if not.

chrisdembia:
Yes this is fixed. Thanks for closing.

PR #431, opened 2017-11-04
[WIP] [appveyor] Set userdir to include commit hash

chrisdembia:


aymanhab:
@chrisdembia Is this related to an open issue? Thanks

chrisdembia:
Yes #416.

chrisdembia:
CI failed because of the updates to opensim-core.

aymanhab:
Thanks, after master finishes building you should be able to merge it into
this branch to get the build to pass, or let me know and I'll do it.

My understanding is that this applies only to ci, so if I build locally
(e.g. to modify icon etc. I still need to make this change manually, or
instead add a target in Ant to do it on demand) is that correct?

Thanks,
-Ayman

On Mon, Nov 6, 2017 at 4:00 PM, Christopher Dembia <notifications@github.com

chrisdembia:
> My understanding is that this applies only to ci, so if I build locally
(e.g. to modify icon etc. I still need to make this change manually, or
instead add a target in Ant to do it on demand) is that correct?

Correct. We can make Ant do it instead.

aymanhab:
Would be wonderful if this is done in Ant but not urgent. Will this PR need redoing then? If so I'll wait, otherwise will merge. Let me know @chrisdembia

chrisdembia:
Okay I'll try doing this in Ant, eventually.

aymanhab:
Wouldn't hurt to merge now due to low risk, changes only on ci. Merging

PR #452, opened 2017-12-05
Detect new version on launch

aymanhab:
On launch, we detect if BuildDate preference is up-to-date, if not we prompt to copy Models to a new directory of user choosing.

### DIFF
--- a/Gui/opensim/utils/src/org/opensim/utils/Bundle.properties
+++ b/Gui/opensim/utils/src/org/opensim/utils/Bundle.properties
@@ -3,3 +3,4 @@ OpenIDE-Module-Long-Description=\
     File managment and low level GUI utilities
 OpenIDE-Module-Name=utils
 OpenIDE-Module-Short-Description=File managment and low level GUI utilities
+CTL_BuildDate=20171204
--- a/Gui/opensim/utils/src/org/opensim/utils/FileUtils.java
+++ b/Gui/opensim/utils/src/org/opensim/utils/FileUtils.java
@@ -31,6 +31,11 @@ import java.awt.Frame;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
 import java.util.StringTokenizer;
 import java.util.Vector;
 import java.util.prefs.Preferences;
@@ -559,5 +564,22 @@ public final class FileUtils {
 	//windows
 	return (os.indexOf( "win" ) >= 0); 
  
+    }
+    public static void copyFiles(final Path sourcePath, final Path targetPath) throws IOException{
+        Files.walkFileTree(sourcePath, new SimpleFileVisitor<Path>() {
+        public FileVisitResult preVisitDirectory(final Path dir,
+                final BasicFileAttributes attrs) throws IOException {
+            Files.createDirectories(targetPath.resolve(sourcePath
+                    .relativize(dir)));
+            return FileVisitResult.CONTINUE;
+        }
+
+        public FileVisitResult visitFile(final Path file,
+                final BasicFileAttributes attrs) throws IOException {
+            Files.copy(file,
+                    targetPath.resolve(sourcePath.relativize(file)));
+            return FileVisitResult.CONTINUE;
+        }
+    });
     }
 }
--- a/Gui/opensim/utils/src/org/opensim/utils/TheApp.java
+++ b/Gui/opensim/utils/src/org/opensim/utils/TheApp.java
@@ -191,7 +191,7 @@ public final class TheApp {
         String userSelection = FileUtils.getInstance().browseForFolder("Folder to install Resources (Models, Scripts):");
         String[] subdirs = new String[]{"Models"}; // Add more folders here as needed
         if (userSelection != null){
-            String src = getInstallDir();
+            String src = getInstallDir()+File.separatorChar+".."+File.separatorChar;
             String dest = userSelection;
             System.out.println("copy resources from "+src+" to "+dest);
             CopyOption[] options = 
--- a/Gui/opensim/utils/src/org/opensim/utils/TheApp.java
+++ b/Gui/opensim/utils/src/org/opensim/utils/TheApp.java
@@ -29,8 +29,11 @@ package org.opensim.utils;
 
 import java.awt.Image;
 import java.io.File;
+import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.CopyOption;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Map;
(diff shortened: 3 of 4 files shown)


chrisdembia:
@aymanhab I tested this out and the "Folder to install Resources" window popped up upon launch :).

Here's some feedback:

- The file dialog opened to my home folder. Ideally it would open to `$HOME/Documents/OpenSim <version>`.
- The dialog did not have the option to create new folders, which should be available if the user wants to create a new directory for storing the files.
screen shot 2017-12-06 at 3 47 16 pm

- Consider changing the title of the dialog from `Folder to install Resources (Models, Scripts):` to `Choose a folder to install models and scripts:`.
- After choosing a folder, nothing was copied to that directory and I got an error:

```
java.lang.NullPointerException
	at org.opensim.utils.FileUtils.browseForFolder(FileUtils.java:277)
	at org.opensim.utils.FileUtils.browseForFolder(FileUtils.java:248)
	at org.opensim.utils.TheApp.installResources(TheApp.java:191)
	at org.opensim.view.Installer.restorePrefs(Installer.java:136)
	at org.opensim.view.Installer.restored(Installer.java:97)
	at org.netbeans.core.startup.NbInstaller.loadCode(NbInstaller.java:468)
[catch] at org.netbeans.core.startup.NbInstaller.loadImpl(NbInstaller.java:391)
	at org.netbeans.core.startup.NbInstaller.access$000(NbInstaller.java:102)
	at org.netbeans.core.startup.NbInstaller$1.run(NbInstaller.java:343)
	at org.openide.filesystems.FileUtil$2.run(FileUtil.java:437)
	at org.openide.filesystems.EventControl.runAtomicAction(EventControl.java:127)
	at org.openide.filesystems.FileSystem.runAtomicAction(FileSystem.java:519)
	at org.openide.filesystems.FileUtil.runAtomicAction(FileUtil.java:421)
	at org.openide.filesystems.FileUtil.runAtomicAction(FileUtil.java:441)
	at org.netbeans.core.startup.NbInstaller.load(NbInstaller.java:340)
	at org.netbeans.ModuleManager.enable(ModuleManager.java:1365)
	at org.netbeans.ModuleManager.enable(ModuleManager.java:1170)
	at org.netbeans.core.startup.ModuleList.installNew(ModuleList.java:340)
	at org.netbeans.core.startup.ModuleList.trigger(ModuleList.java:276)
	at org.netbeans.core.startup.ModuleSystem.restore(ModuleSystem.java:301)
	at org.netbeans.core.startup.Main.getModuleSystem(Main.java:181)
	at org.netbeans.core.startup.Main.getModuleSystem(Main.java:150)
	at org.netbeans.core.startup.Main.start(Main.java:307)
	at org.netbeans.core.startup.TopThreadGroup.run(TopThreadGroup.java:123)
	at java.lang.Thread.run(Thread.java:745)
```
- I loaded the GUI a second time and I still got the pop-up, and upon selecting a folder, I got the same error.

chrisdembia:
Some more comments:

- We no longer need the symlinks for Models, Scripts, etc.

aymanhab:
Thanks for trying this out Chris. The folder name you suggested can't be
used as default since it doesn't exist yet. On windows I was able to create
a folder or just enter the name  and have the code create it. Don't know
the difference on osx. If no exception is thrown it should set the build
date correctly and you wouldn't see the prompt again. Will continue
testing, thanks for the feedback.


On Wed, Dec 6, 2017 at 3:52 PM, Christopher Dembia <notifications@github.com

chrisdembia:
>  The folder name you suggested can't be
used as default since it doesn't exist yet.

Hmm...what about a text field then instead of a file dialog? The dialog with a textbox could also have a button "Select existing folder" that would bring up a file dialog.

aymanhab:
@chrisdembia I agree with the title/prompt change so I did it. Thanks.
There's a mode for FileBrowser to allow for creating new folders, I enabled that too.
I still prefer a FileBrowser over a text field that is more error prone, with spaces, tabs, \ etc. We get the checking for free by using the FileBrowser. The main issue left is locating the Models folder within the installation which is platform dependent.

aymanhab:
I thought so too but it's hard to verify since the build/dev environment layout is different from installed one. 

I like the three option dialog but it opens the door for other issues as well e.g what if the user agrees to Documents/OpenSim 4.0 but he/she has no access or the folder already exists? You may feel better about the plain directory browser with default if allowed to create new folders right there. Just pushed this change. May need to merge to get the great installer auto-build.

chrisdembia:
> I like the three option dialog but it opens the door for other issues as well e.g what if the user agrees to Documents/OpenSim 4.0 but he/she has no access or the folder already exists? 

Yes, I see this gets complicated, but most users will want to just use the default. These conditions could be checked before the dialog comes up. I see it as very unlikely though that the user won't have access to that folder. And if the folder already exists, the dialog could say "Warning: this folder already exists and will be overwritten if you select OK."

aymanhab:
Ok, I implemented the three options dialog and tested on OSX. The changes were merged into master so should appear in next automated build. Please let me know how that works (at least the default behavior on fresh machine as it will be first user experience on osx)

chrisdembia:
Thanks @aymanhab it works much better now. I really like how "Open Model..." opens to the resources dir! I have some suggestions and I would be willing to submit a PR for them. 

- In my opinion, the version number in the suggested folder name should match exactly the `app.version` property in Ant.
- Is storing the BuildDate in Preferences a good idea? It's not really a preference; that is, why would a user want to see this or change it?
- If I click Customize and then Cancel, I expect to return to the original dialog box. Instead, the GUI loads and I get an exception message (red circle).
- I think we could come up with a better button name than Customize, because one can customize simply by editing the text field. Perhaps just put a button after the text field with either a folder icon or 3 dots? A similar UI is used in the Tool windows for "open file dialog."

aymanhab:
Chris,

Good points, thanks for the feedback..

1. I agree but was not sure how to get this number inside the application.
I don't think it needs to change with every build/release and was not sure
when app.version would change/update..
2. Storing BuildDate as a preference is a great way for us to detect first
launch and to test re-installing resources (by changing the number to a
smaller one manually). There are many preferences that are hidden from the
dialog so once we're done testing/tweaking we should hide it indeed.
3. Both customize/Cancel and just cancel will lead to problems downstream
with unwritable or hard to reach folders. Not sure I like having the dialog
stay on  either. Not a big issue since this is not default behavior but
feel free to PR or suggest alternatives.
4. I think I took "Customize" from your "Select Custom Folder" earlier
(since it's shorter and about same size as other buttons), we could use
Browse as well or make the selection similar to other places in the GUI
that browse for a folder.

chrisdembia:
> 1. I agree but was not sure how to get this number inside the application.
I don't think it needs to change with every build/release and was not sure
when app.version would change/update..

Hmm...maybe it would be good to have a "OPENSIM_VERSION_MAJOR_MINOR" stored in one place. Ideally this could be obtained by pulling off the major and minor versions from `app.version` to avoid duplication beyond `app.version`.

> 4. I think I took "Customize" from your "Select Custom Folder" earlier
(since it's shorter and about same size as other buttons), we could use
Browse as well or make the selection similar to other places in the GUI
that browse for a folder.

Yes I understand, but I originally thought the text field would not be editable. Since it is editable, users can customize simply by editing the text field, without clicking the "Customize" button. I like "Browse"; I certainly agree with being consistent across the UI.

PR #456, opened 2017-12-07
On Mac, allow installing multiple versions.

### LINKED ISSUE #426: [Mac] Installer overwrites pre-existing OpenSim installation
We will need to address this so that users can have OpenSim 4.0 and OpenSim 4.1 etc. installed at the same time. The name of the app might need to contain the version number.

chrisdembia:
Without this change, Installer.app would delete previous versions of the app on the user's machine. By giving the pkg a unique identifier for each version, we no longer have that issue!

Fixes #426

### DIFF
--- a/Gui/opensim/installer_files/OSX/Info.plist
+++ b/Gui/opensim/installer_files/OSX/Info.plist
@@ -17,6 +17,9 @@
     
     <key>CFBundleShortVersionString</key>
     <string>@VERSION@</string>
+
+    <key>CFBundleIdentifier</key>
+    <string>org.opensim.app.@VERSION@</string>
     
     <key>CFBundleSignature</key>
     <string>????</string>
--- a/Installer/macOS/Distribution.xml
+++ b/Installer/macOS/Distribution.xml
@@ -2,7 +2,7 @@
 <installer-gui-script minSpecVersion="1">
     <title>OpenSim</title>
     <background file="OpenSimLogoWhiteNoText.png" scaling="none" alignment="bottomleft"/>
-    <pkg-ref id="org.opensim.app.pkg"/>
+    <pkg-ref id="org.opensim.app.pkg.@VERSION@"/>
     <!--Users will need to choose the Home directory until we stop writing JSON
          files within the app. 
          https://stackoverflow.com/questions/12863944/how-do-you-specify-a-default-install-location-to-home-with-pkgbuild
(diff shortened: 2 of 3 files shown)


aymanhab:
Changes look reasonable but boils down to:
1. Whether this has been tested locally.
2. Interaction between versions in terms of Preferences, and resource files (artifacts that do not live with the installed application).
If that seems sufficiently tested, I'd be happy to merge.

chrisdembia:
I tested this locally briefly. The preferences are not versioned, so all versions would share the same preferences.

aymanhab:
With Preference shared there will be side effects (same on windows) e.g. Working directories, other visual preferences, Geometry search path etc. If you think this is helpful for internal usres/testing on osx (since the solution is osx specific) then we can merge but I'm not convinced this is something we would advertise. Do you see this used for having 4.0 and 4.x on same machine, or multiple beta builds of 4.0? If the former than we maybe able to solve the Preferences issues by having version specific preferences.

chrisdembia:
>  Do you see this used for having 4.0 and 4.x on same machine, or multiple beta builds of 4.0? 

Both (even having 4.x and 5.x side by side). It was common for me to have multiple versions of 3.x installed simultaneously on Windows. How were preferences handled there?

Yes I think having version-specific preferences would work. I even think it would be fine if we do not inherit preferences from a previous install; it's not too much of a burden for users to manually change them again (they might even want a clean slate). I'm not sure how to implement versions-specific preferences; the choice of identifier for preferences (`org.opensim.utils`) might be made by NetBeans.

aymanhab:
@chrisdembia I like the  version specific preferences for minor versions (4.0, 4.1, ...5.0) but not for every patch/build as it will bloat the storage (Registry on Windows, FileSystem on osx). The Preferences are associated with a class in org.opensim.utils rather than selected by Netbeans, I could make a version specific class and use that for preferences but that will make it harder to detect/carry-over older preferences and will make it clunkier to upgrade. Do you think it's worth it?
On windows multiple installation shared preferences in the past (all stored/retrieved from Registry)

chrisdembia:
>  I like the version specific preferences for minor versions (4.0, 4.1, ...5.0) but not for every patch/build as it will bloat the storage (Registry on Windows, FileSystem on osx).

I agree.

Ideally the Java code would pull the version number from the Ant property. Thanks for explaining the preferences; is there a way to override the identifier used for that class? I do *not* think we should make a class for each version.

Given that, on WIndows, preferences were shared in the past, I think having version-specific preferences is a separate issue from allowing parity with what we can already do on Windows (multiple simultaneous installations; this PR). If having version-specific preferences is hard to implement for this PR, we could open a new issue.

aymanhab:
Sounds reasonable, please open an issue for the version specific preferences and merge this one. Thank you @chrisdembia

chrisdembia:
I opened #460

aymanhab:
Failures due to not finding java, unrelated. Merging

chrisdembia:
Thanks @aymanhab

PR #458, opened 2017-12-09
Update models and threejs

aymanhab:
Update threejs and opensim-models commits used on GUI master


aymanhab:
Updating to latest version so automated builds are up to date.

Issue #472, opened 2018-01-11
Hidden bodies are still hidden after save/re-open, but hidden muscles appear again


tkuchida:
- open a model
- hide all bodies and muscles
- save, close, re-open

Bodies do not appear, but muscles are shown.

tkuchida:
Verified fixed in AppVeyor artifact OpenSim-2fd23ef3-2018-02-20.zip :checkered_flag:

PR #480, opened 2018-01-11
Create non "visible" geometry

aymanhab:
Even if geometry is not currently marked as "visible", we now create it so that if the user changes his/her mind during GUI session, toggling of visibility is not made complicated/expensive. This also fixes the issue where Hidden muscles are restored on deserialization.


tkuchida:
"the issue where Hidden muscles are restored on deserialization" was #472.

PR #500, opened 2018-02-02
[WIP] [appveyor] Prep for opensim-core-VC141

chrisdembia:
WIP; don't merge yet.

PR #503, opened 2018-02-05
Upgrade API to latest on master. Main changes OutputSpatialVec, Table…

aymanhab:
…ReporterSpatialVec, "InvalidScaleFactors_get"


aymanhab:
This upgrades the Java bindings used by the GUI to latest API on master. It's great to see the diffs to understand the effects of API changes on Matlab, Java bindings.  Can somebody clarify why all Components and iterators have a "InvalidScaleFactors_get" method? Is this implemented differently by these classes? Does it need to be public? Thanks

aymanhab:
@chrisdembia This should take care of issue #500

chrisdembia:
`InvalidScaleFactors` is public static const member variable in ModelComponent, introduced for scaling. We should probably hide it from SWIG. I created https://github.com/opensim-org/opensim-core/issues/2074.

I will merge this now.

PR #504, opened 2018-02-07
Rearrange installation layout on Mac; resources into single Resources dir.

### LINKED ISSUE #502: Mac: Consider placing lib, include symlinks in the root of the install folder instead of in sdk
Right now, the Mac app places symlinks for the lib and include folders in the sdk folder in the installation. This means that when using CMake, users must actually specify `<install-dir>/sdk/`, which is different than on Windows. I think we should consider placing these symlinks in the root of the install folder, especially since the `APIExamples` are now installed in the Documents folder.

Relatedly, we should no longer create symlinks for the APIExamples or Scripts folders.

chrisdembia:
This PR cannot be merged until https://github.com/opensim-org/opensim-core/pull/2079 is merged and its CI tests are finished.

With this PR, the resources are rearranged into a single Resources directory in the GUI that can be copied over to the user's Documents folder. We also use an updated organization of code examples from the API.

This PR also reorganizes the installation layout on Mac (reduces the number of symlinks, and introduces a top-level cmake symlink).

I tested that, on Mac, the first launch of the GUI causes the resources to be installed to my user directory, with the new layout of code examples from the API.

I also tested, on Mac, that I can compile a CMake C++ API example using the API that comes with the GUI, specifying the root of the install layout as the CMAKE_PREFIX_PATH (fixes #502).

Mac installation layout:
screen shot 2018-02-07 at 2 36 22 pm

Mac resources layout (needs to be fixed):
screen shot 2018-02-07 at 2 49 53 pm

TODO: 

- In TheApp.java, the entire Resources directory is copied into the
Documents folder, rather than just the contents of the Resources
directory. Editing the code in TheApp.java to achieve the desired behavior did not seem trivial for me.
- On Windows, I'm not sure that the GUI_Scripting folder is installed.
- On Windows, I'm not sure that the API's Resources folder is copied over correctly.
- On Mac, should we symlink Geometry to the root of the installation (currently, we do)?
- On both platforms, the resources should not be accessible from the installation (at the UEFS this Monday, some people were confused between the models in Documents and in the GUI folder). I suggest that the Ant tasks zip up the Resources folder.

@aymanhab can you look into the TODOs?


aymanhab:
Thanks @chrisdembia 
- TheApp.java has a list of folders relative to top level that get copied, line 240 has srcPath specification so we can append "Resources" there if needed. Relative paths are listed in subdirs[]. I can do it after all other changes are done.
- GUI_Scripting doesn't make it to the installation on Windows, used to be under top level Script sfolder in 3.3. Will fix after all your changes are done.
- I'm ok with symlinking Geometry since ti's much harder to dig into the .app folder on osx

chrisdembia:
> TheApp.java has a list of folders relative to top level that get copied, line 240 has srcPath specification so we can append "Resources" there if needed. Relative paths are listed in subdirs[]. I can do it after all other changes are done.

Yes but I don't think we want to enumerate the directories within `Resources`. The idea here was that we could copy all its contents.

chrisdembia:
@aymanhab https://github.com/opensim-org/opensim-core/pull/2079 is now merged, which means this can be worked on

I added a TODO item above: On both platforms, the resources should not be accessible from the installation (at the UEFS this Monday, some people were confused between the models in Documents and in the GUI folder). I suggest that the Ant tasks zip up the Resources folder.

I'm happy to help a little with these TODOs but I don't think I can do them all myself (some of it involves using Java to unzip the resources folder). Perhaps we should chat again.

chrisdembia:
@aymanhab I pushed some commits in which the Resources folder is now a ZIP in the distribution, and "installing resources" now unzips that ZIP.

I tested that this works on Mac, but I did not check error cases.

I also tried to add copying of GUI scripts on Windows and of zipping resources on Windows, but I did not do any testing on Windows.

aymanhab:
Tested on windows from dev environment (different layout). When .zip file is found all worked perfectly. Need to trigger on new installation by changing resource bundle for TheApp.class, "CTL_BuildDate"

chrisdembia:
> Need to trigger on new installation by changing resource bundle for TheApp.class, "CTL_BuildDate"

Is this for you or for me?

aymanhab:
I did it but wanted to leave a record of how we force the reinstallation of resources for future use.Also for testing purposes, you can manually invoke the installation of resources using installResources() call from the scripting shell.

chrisdembia:
Okay cool. That `installResources()` command is neat.

aymanhab:
@chrisdembia is this ready to merge?

chrisdembia:
@aymanhab before this is merged, it would be good if you could ensure that on Windows, GUI scripts are installed into the resources dir, and if you could test some error cases (e.g., if the zip is missing, if the destination folder already exists, etc.)

aymanhab:
I did check that GUI scripts are installed correctly. The dialog used to check the directory doesn't exist before copying, did that change?

chrisdembia:
> The dialog used to check the directory doesn't exist before copying, did that change?

I don't know. I modified code in TheApp without fully understanding the reprecussions.

aymanhab:
As expected, the dialog didn't allow installation into existing folder, once new folder is selected it all worked fine. LGTM.

aymanhab:
@chrisdembia Any reason not to merge this after your testing and mine? Thanks

chrisdembia:
Not that I can think of!

Issue #163, opened 2017-06-07
Fix issues with how marker and grf data display on open/close


jenhicks:

jenhicks:
Closing this issue as persistence and zoomed in are addressed in other issues.

PR #507, opened 2018-02-09
Fix bounding box on preview markers

aymanhab:
Send bounding box to visualizer so that camera zooms out on load.


aseth1:
Which issue #s does this PR address?

aymanhab:
Address issue #163

PR #515, opened 2018-02-12
Restore scope of  methods to protected for proxy classes

aymanhab:
No need to make internal methods/constructors public in Proxy classes to avoid clutter


aymanhab:
Once this builds, will submit pr to opensim-core to fix the generation of Java files from Cpp

aymanhab:
This actually fails because there're a couple dozen uses of getCPtr outside the modeling package primarily to check for equality. While these may be changed to another test, it seems outside the scope so will close this PR and make no changes to opensim-core.

Just FYI @chrisdembia Thanks for your help.

chrisdembia:
Okay.

PR #526, opened 2018-02-22
Add template for GitHub Issues

tkuchida:


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
+### Steps to reproduce
+
+### Expected result
+
+### Actual result (include screen grab where appropriate)
```
chrisdembia: If you put `(include screen grab where appropriate)` on its own line, it will be easier to replace this placeholder text with an actual screen grab.

tkuchida: Updated—though it now looks like replacing the parenthetic comment with a screen grab is optional when the actual result should always be specified.

PR #532, opened 2018-02-24
Expose the method available in TheApp class to installResources 

aymanhab:
Allow users to manually reinstall resources


chrisdembia:
Neat!

Issue #556, opened 2018-02-26
Probes not shown in Navigator


tkuchida:
### Steps to reproduce
- open subject01_metabolics.osim from [Simulation-Based Design to Reduce Metabolic Cost](https://simtk-confluence.stanford.edu/display/OpenSim/Simulation-Based+Design+to+Reduce+Metabolic+Cost)

### Expected result
*Probes* -> *metabolics* shown in Navigator.

### Actual result
*Probes* does not appear.

### Environment and GUI version
OpenSim-2416f17a-2018-02-25.zip ("OpenSim 4.0 User Experience and Feedback Session" version)

aymanhab:
Addressed by PR #573 and PR #565

jenhicks:
Fixed on windows OpenSim-e7831221-2018-03-06

PR #565, opened 2018-02-27
Add node for Probes in navigator

aymanhab:
Fix missing Probes node in navigator issue #556

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/NavigatorByTypeModel.java
+++ b/Gui/opensim/view/src/org/opensim/view/NavigatorByTypeModel.java
@@ -52,7 +52,8 @@ import org.opensim.modeling.Marker;
 import org.opensim.modeling.MarkerSet;
 import org.opensim.modeling.Model;
 import org.opensim.modeling.ObjectGroup;
-import org.opensim.modeling.OpenSimObject;
+import org.opensim.modeling.Probe;
+import org.opensim.modeling.ProbeSet;
 
 /**
  *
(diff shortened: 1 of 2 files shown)


tkuchida:
Is it possible to populate the GUI Navigator by iterating through all Components in the Model, filtering out and categorizing bodies, forces, etc., and dropping all remaining Components into a catchall "Other" category?

aymanhab:
Tom, that's exactly what we do but we don't collect the extras in a bucket
because it would be huge and rather hard to manage though we can try.
You'll be surprised what we find there ☺

tkuchida:
> that's exactly what we do but we don't collect the extras in a bucket because it would be huge and rather hard to manage though we can try.

I think the extras should be collected and displayed; otherwise, it sounds like a many-to-one mapping is being performed, where the user would be unable to distinguish between two distinct models once they are opened in the GUI. It would be nice if the "extras" were displayed under an "Uncategorized" label, and if the user had a way of creating new categories (e.g., specify the classname and label in a config file).

aymanhab:
For TugOfWar model:
capturetowarother

tkuchida:
Cool! I think it might be helpful to see these so nothing in the model is hidden from the user. Is it possible to group by concrete class name and alphabetize? Thoughts, @aseth1?

chrisdembia:
> Is it possible to group by concrete class name and alphabetize? 

I like this idea a lot, but I also think this enhancement could be delayed to the next minor release.

aseth1:
I do think it is important to be able to navigate through everything in a `Model` not just a subset of   "convenient" sets. I would like *Other Components* to be the other immediate subcomponents of the `Model`. I then would want each Component to have *Other Components* or *subcomponents* so all Components could be traversed through the navigator. I am not a big fan of pulling all components, in a potentially deep tree, into a flat list at the top-level.

aymanhab:
@aseth1 Is there an interface to get the immediate list of children? (we can traverse all components and check for parent but that seems backwards), if so we can try it and see what we get.   I also don't know if end users of the GUI would care about whether Coordinate is implemented as a Component or not or where any component lie in the tree.

aseth1:
@aymanhab we have:
```cpp
std::vector<SimTK::ReferencePtr<const Component>> getImmediateSubcomponents() const;
``` 
on `Component`. It is currently protected but this could be a case to make it public.

chrisdembia:
I think we should have some function like that that is public.

Ideally the vector would not be created on the fly, and ideally the return type would be `ComponentList` or something like that.

PR #566, opened 2018-02-27
Turn off recording accelerations when using Kinematics analysis in GUI

aymanhab:
When using kinematics internally to record states, no need to compute accelerations or realize higher then velocity.


chrisdembia:
This would seem to be a change in behavior; users may have expected to get accelerations from RRA and now will not. My preference is to call `realizeAcceleration()` within the Kinematics reporter.  In fact, even with this PR, there would be a bug in opensim-core.

chrisdembia:
If this kinematics analysis is only used internally by the GUI (this isn't the kinematics analysis added by RRA/CMC to print kinematics STO files?), then this change is good. However, a bug likely still remains in opensim-core.

aymanhab:
Thanks for the quick turnaround @chrisdembia Yes this kinematics analysis is used to create the animations when running tools in the GUI so there's no user expectation of accelerations.

aseth1:
@aymanhab IMHO the GUI should no longer be using Analyses internally. Use the `StatesTrajectory` and `StatesTrajectoryReporter` for all/any recording and playback of states. Analyses should remain only for user legacy reasons.

Issue #511, opened 2018-02-10
The "running dude" button generates simulations of inconsistent duration


tkuchida:
**Steps to reproduce:**
1. Open a model and set the :running_man: *end time* to 5.0.
2. Click :running_man: and play the animation.
3. Repeat step 2.

**Results:**
Simulations are 4.993, 4.999, and 4.998 seconds long.

![](https://user-images.githubusercontent.com/4203505/36067523-726d655e-0e73-11e8-9085-e7784dab5505.png)
![](https://user-images.githubusercontent.com/4203505/36067524-7579b040-0e73-11e8-88c6-4e79aa1bd97e.png)
![](https://user-images.githubusercontent.com/4203505/36067527-7785601e-0e73-11e8-9c9d-410b34fa3f8f.png)

**Environment:**
*OpenSim-078b8daf-2018-01-26.zip* (obtained from AppVeyor artifacts) on Windows.

chrisdembia:
This might be related to the state that the Manager returns, which may have been changed in the improvements to Manager.

chrisdembia:
Jenny has this issue too with the Tug Of War model and the Forward Tool. The final time in the forward tool was 0.5 but the simulation only went to 0.48 seconds. So, this is not only an issue with the running dude.

UEFS

aymanhab:
I believe this is fixed by PR #571 Please verify when it's merged

aymanhab:
Since PR #571 has been merged, this should be fixed in latest build.

jenhicks:
This is fixed for me Windows 10 OpenSim-e7831221-2018-03-06

tkuchida:
Verified using :running_man: button and Forward Tool. AppVeyor artifact OpenSim-e7831221-2018-03-06.zip.

tkuchida:
Reopening because @jenhicks tested using the same AppVeyor artifact and I presume wants this tested on mac also?

PR #571, opened 2018-03-01
Last frame was missing from tools because step() is not called

aymanhab:
Change in core caused Step not to be called on last frame, instead end is called. Update the animation (analysis) on GUI side to follow this convention.

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/motions/JavaMotionDisplayerCallback.java
+++ b/Gui/opensim/view/src/org/opensim/view/motions/JavaMotionDisplayerCallback.java
@@ -347,6 +347,7 @@ public class JavaMotionDisplayerCallback extends AnalysisWrapperWithTimer {
     }
    @Override
     public int end(State s) {
+        processStep(s, storage.getSize());
         return  super.end(s);
     }
 }
--- a/Gui/opensim/view/src/org/opensim/view/motions/JavaMotionDisplayerCallback.java
+++ b/Gui/opensim/view/src/org/opensim/view/motions/JavaMotionDisplayerCallback.java
@@ -347,7 +347,10 @@ public class JavaMotionDisplayerCallback extends AnalysisWrapperWithTimer {
     }
    @Override
     public int end(State s) {
-        processStep(s, storage.getSize());
+        if (kinReporter != null)
+            kinReporter.end(s);
+        if (statesReporter!=null)
+            statesReporter.end(s);
         return  super.end(s);
     }
 }


aymanhab:
Address issue #511

aseth1:
@aymanhab AFAICT in core, calling `Analysis::end(s)` is sufficient to record the final state correctly. In no instances are we calling `step(s)` in `end(s)`. Therefore if your `Analysis` is calling `processStep(s, N)` (e.g. `step(s, N)`)  in `end(s)` it is definitely inconsistent with Analyses in the API.  If you look at all the `end(s)` implementations they just call `record(s)` as does `step(s)` so I can see why you are needing to call `processStep()` which is likely is the recording guts. I would rename it `record(s)` and have your `step()` and `end()` call it. Then it will at least be consistent with the `Analyses` in core.

There was also an inconsistency between the `AnalyzeTool` and `ForwardTool` (Manager) and this was resolved by only calling `step(s)` during our march through states in `AnalyzeTool::run()` and integration and only `end(s)`, when reaching the last index of input data (`AnalyzeTool`) or finished integrating in the case of the `Manager`, and never `step(s)` and `end(s)` together.

aymanhab:
@aseth1 thanks for the pointers as to why the behavior was changed. The "Analysis" being modified here is what records the kinematics or states so that they can be played back when tools are running. Since all these tools are "Analysis" based, the GUI follows suit (internally using StatesReporter or KinematicsReporter)  both of these invoke record (which is not part of the interface by the way) inside their end method. The GUI based Analysis uses the signature that takes a step number of "step" so that it can report progress. Please let me know if you have any further questions or concerns regarding this change.

aseth1:
@aymanhab I know `record(s)` is not part of the public interface. The problem is that the public virtual methods `begin(s)`, `step(s, N)` and `end(s)` all have to do the same operation of recording something plus doing something else (like initialize Storage or finalize something). I do not like or approve of this interface - it is what it is. It appears that `record(s)` was invented to support the common operation that all three public methods have to do.

I think it safest moving forward to use a similar structure and not call `step(s)` or in your case `processStep(s)` inside of `end(s)` because if anything has to be added to `step()` you will inadvertently also bake it into `end(s)`. I am not saying that is happening, it just exposes your code to that risk and that may be hard to track down and debug in the future. 5 min know could save the team hours or even days hunting months and years down the road.

aymanhab:
Thanks for the feedback @aseth1 I made a change to delegate the call to end(s) in the GUI Analysis to the corresponding end(s), and checked that results files saved in GUI have correct final frame and that it shows exactly once. Please let me know if you have any further concerns.

aymanhab:
Thanks @aseth1 Merging since build failures on travis are due to sourceforge issues.

PR #573, opened 2018-03-02
Use enabled property for probes

aymanhab:
Switching to new name for "Enabling" property for Probes, tested in GUI.


aymanhab:
Thanks @aseth :+1:

Issue #534, opened 2018-02-25
UnsupportedOperationException after entering coordinate value outside clamp range


tkuchida:
### Steps to reproduce
- open gait10dof18musc.osim
- in Navigator > Coordinates, type "-100" in the *Value* textbox for *pelvis_tilt* and press `<Enter>`

### Expected result
- *pelvis_tilt* changes to -100° in visualizer and "-100" is shown in *Value* textbox

### Actual result
Red-circle error:
<pre>
java.lang.UnsupportedOperationException
	at org.opensim.coordinateviewer.CoordinateSliderWithBox$handleReturnAction.focusLost(CoordinateSliderWithBox.java:575)
	at java.awt.AWTEventMulticaster.focusLost(AWTEventMulticaster.java:230)
	at java.awt.Component.processFocusEvent(Component.java:6425)
	at javax.swing.JFormattedTextField.processFocusEvent(JFormattedTextField.java:612)
	at java.awt.Component.processEvent(Component.java:6289)
	at java.awt.Container.processEvent(Container.java:2237)
	at java.awt.Component.dispatchEventImpl(Component.java:4889)
	at java.awt.Container.dispatchEventImpl(Container.java:2295)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.KeyboardFocusManager.redispatchEvent(KeyboardFocusManager.java:1954)
	at java.awt.DefaultKeyboardFocusManager.typeAheadAssertions(DefaultKeyboardFocusManager.java:995)
	at java.awt.DefaultKeyboardFocusManager.dispatchEvent(DefaultKeyboardFocusManager.java:661)
	at java.awt.Component.dispatchEventImpl(Component.java:4760)
	at java.awt.Container.dispatchEventImpl(Container.java:2295)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80)
[...]

PR #584, opened 2018-03-10
Fix two unhandled exceptions

aymanhab:
When editing coordinates and when cancelling out from color selection dialog


aymanhab:
Address #534 and unreported exception due to cancelling of Color selection dialog

Issue #542, opened 2018-02-26
Crash when adding and moving new marker


tkuchida:
### Steps to reproduce
- open a model
- in *Navigator*, right-click *Markers* and select *Add New*
- hold `<Ctrl>` and left-click marker circle in Visualizer window

### Expected result
Marker's axes appear, allowing user to drag Marker around visualizer.

### Actual result
Marker's axes automatically appear, disappear, appear, disappear, … (approx. 2 cycles/second), and a list of the text "Marker:NewMarker_2" fills the status bar.

![](https://user-images.githubusercontent.com/4203505/36660076-9a79f64c-1a8b-11e8-9aed-43b0a6f53ba3.png)

### Environment and GUI version
OpenSim-2416f17a-2018-02-25.zip ("OpenSim 4.0 User Experience and Feedback Session" version)

tkuchida:
Also occurs just by clicking the marker (i.e., no `<Ctrl>`). Marker is selected after first click, but immediately becomes deselected. After clicking again, the GUI enters a loop of selecting and deselecting the marker.

jenhicks:
This seems to be working, except for the first marker that I add is bigger than subsequent markers. 
Load gait2354
Add marker
Move it around (it gets bigger)
Add a second marker 
Move it around (it doesn't get bigger)

chrisdembia:
At the UEFS, Hannah had an issue where the GUI autonomously toggled between two markers in the navigator at 1 or 2 cycles per second.

aymanhab:
@chrisdembia indeed that was an issue due to circular notifications being sent between the navigator and visualizer but that was fixed (by utilizing a master list of selected objects as was done in 3.3 and earlier).

jimmyDunne:
Markers can be added and selected without crashes. However, when you try to move it the marker gets scaled- I have opened a new issue for this at #695

Issue #575, opened 2018-03-05
Double-clicking bodies in "topology view" works only once per body


tkuchida:
### Steps to reproduce
- open arm26 model
- *Window* menu > *Topology View*
- double-click "r_humerus" box -> "r_humerus" body is selected in the Navigator, as expected
- double-click "r_ulna_radius_hand" box -> "r_ulna_radius_hand" body is selected in the Navigator, as expected
- double-click "r_humerus" box

### Expected result
"r_humerus" body is selected in the Navigator

### Actual result
Nothing happens. If the model is then closed, the GUI reports a red-circle error (`NullPointerException`) when the model is opened again:
<pre>
java.lang.NullPointerException
	at org.opensim.view.pub.ViewDB.markSelected(ViewDB.java:1022)
	at org.opensim.view.pub.ViewDB.clearSelectedObjects(ViewDB.java:1159)
	at org.opensim.view.pub.ViewDB.update(ViewDB.java:340)
	at java.util.Observable.notifyObservers(Observable.java:159)
	at org.opensim.view.pub.OpenSimDB.setCurrentModel(OpenSimDB.java:264)
	at org.opensim.view.pub.OpenSimDB.addModel(OpenSimDB.java:156)
	at org.opensim.view.pub.OpenSimDB.addModel(OpenSimDB.java:143)
	at org.opensim.view.FileOpenOsimModelAction$2.run(FileOpenOsimModelAction.java:98)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:726)
	at org.netbeans.core.TimableEventQueue.dispatchEvent(TimableEventQueue.java:159)
[catch] at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
[...]

PR #593, opened 2018-03-14
Fix circular notification in select

aymanhab:
Keep track of SelectedObjects in GUI and Visualizer to avoid sending unnecessary/circular notifications between them and to cleanup properly on close.


aymanhab:
Fixes issues #542, #575

PR #596, opened 2018-03-16
Make windows installer

aymanhab:
Starting point for Windows Installer using NSIS, as of now it only does zip/unzip and creates Desktop shortcut.


aymanhab:
To integrate with ci need to copy the script to the top level folder "dist/installer" then run NSIS on it. Would you have some cycles to help integrate @chrisdembia ? Thanks

aymanhab:
Integration into ci could happen later. Need to merge so I can test/make installers on different Windows machines.

chrisdembia:
I'm happy to help at some point.

Issue #580, opened 2018-03-10
User is prompted for desired location of models and scripts directory, but GUI does not copy them there and does not communicate this to the user


tkuchida:
### Steps to reproduce
- start OpenSim for the first time
- when prompted, specify desired directory for models and scripts

### Expected result
Models and scripts are installed to specified directory and the user is informed that the operation was successful.

### Actual result
Two dialog boxes appeared; neither was helpful. The *Models* directory contains models as expected, but the *sdk* directory contains only two empty directories: *APIExamples* and *Scripts*. I presume there are supposed to be API examples in the *APIExamples* directory, but there aren't any. The user isn't told whether the operation completed as expected or whether a something went wrong.

First dialog box:
![](https://user-images.githubusercontent.com/4203505/37240455-72968d74-2400-11e8-8538-049cf4f95ec0.png)
Second dialog box:
![](https://user-images.githubusercontent.com/4203505/37240461-7c8589e8-2400-11e8-8430-687ff985df8c.png)

Contents of *Messages* window:
<pre>
Copying resources to C:\Users\tkuchida\OpenSim 4.0 appveyor Mar 6 2018\Models and Scripts
</pre>

### Environment and GUI version
AppVeyor artifact OpenSim-e7831221-2018-03-06.zip

jenhicks:
@aymanhab Is this new to 4.0?

aseth1:
Yes. We now copy editable "resources" which include the Models and CodeExamples (instead of sdk) into the Users home space. This is typically:
`C:\Users\<user_name>\Documents\OpenSim`

aymanhab:
Tom, we replaced this part with a zip/unzip mechanism, can you report status on latest appveyor build? Thanks. You should be able to reinvoke the process by typing installResources () in the scripting shell.

aymanhab:
@chrisdembia The Resources.zip on the windows AppVeyor build ends up with CodeExamples but no Models. Any idea what might be different from the version I tested locally? Thanks

tkuchida:
> Tom, we replaced this part with a zip/unzip mechanism, can you report status on latest appveyor build?

[...]

PR #600, opened 2018-03-18
Copy Models before building Resources.zip as part of PrepareInstaller

aymanhab:
AppVeyor build scripts copies Models after Resource.zip has been created. This PR switches the order so that Models are copied first and so included in Resources.zip.


aymanhab:
Fixes issue #580

aymanhab:
Another option is to use the target ALL_BUILD instead of PrepareInstaller.

aymanhab:
Thanks @chrisdembia

PR #608, opened 2018-03-20
Use Model.formStateStorage in MotionPlayer

aymanhab:
This should speed up playback of old result files that are not recognized as states. It delegates label/order gymnastics to API.  Need to test:
- [x] Saved output from tools don't have extraneous states
- [x] Preview of Experimental data is not affected
- [x] Speedup realized (IK, SO, other tools)

### DIFF
--- a/Gui/opensim/tracking/src/org/opensim/tracking/IKToolModel.java
+++ b/Gui/opensim/tracking/src/org/opensim/tracking/IKToolModel.java
@@ -160,11 +160,9 @@ public class IKToolModel extends Observable implements Observer {
                 }
             }
             // Create a new States storage and load it in GUI 
-            // this should make playback faster since no need to assemble
-            Storage ikmotion = new Storage(512, "IKResults"); // Java-side copy
-            getOriginalModel().formStateStorage(
-                    animationCallback.getStorage(),
-                    ikmotion, false);
+            // We don't create full state to avoid polluting results
+            Storage ikmotion = new Storage(animationCallback.getStorage()); // Java-side copy
+            ikmotion.setName("IKResults");
             updateMotion(ikmotion);
 
             setExecuting(false);
--- a/Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
+++ b/Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
@@ -393,6 +393,10 @@ public class MotionDisplayer {
         if (simmMotionData == null)
            return;
 
+        // We create a temporary array to hold interpolated values, in case Slider doesn't coincide with data
+        colNames = simmMotionData.getColumnLabels();
+        int numColumnsIncludingTime = colNames.getSize();
+        interpolatedStates = new ArrayDouble(0.0, numColumnsIncludingTime-1);
         // If provided simmMotionData is empty or have one frame, then we're building it live and can't
         // convert to full State form, will keep using old mapping until Tool run is finished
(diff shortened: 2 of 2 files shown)


aymanhab:
@aseth1 This is mostly working, need to test the items in the top comment but speedup should be already in. Would be good to verify that part early.

aseth1:
@aymanhab I noticed a sizeable speedup and much less CPU time associated with apply the time and state values to the model! Motion was correct as well. I profiled for good measure:
image

An you can see that applying the new time (an related state values) to the model is now only 12.5% of total CPU down from over 80%.

aymanhab:
Should we leave behind the console message 
"Number of columns does not match in formStateStorage. Found 37 Expected  110."?

aymanhab:
When running SO, should the resulting motion in the navigator contain CoordinatesOnly or full state? Ideally we get an actual state with SO activations but that seems outside the scope.

aymanhab:
Pending feedback and changes based on items above, this is ready.

aseth1:
> When running SO, should the resulting motion in the navigator contain CoordinatesOnly or full state? Ideally we get an actual state with SO activations but that seems outside the scope.

It depends. If I load a motion (from IK) and then run SO, I would not want the motion I loaded to be modified. If the AnalyzeTool (i.e. SO) generated *SO-Results*, I would want those results to include the activations that SO computed so that I could save and replay the SO-Results identically to what was generated when I ran SO live.

aymanhab:
Since we want others to test animation performance and  since proposed changes to SO functionality/playback is not within the scope of this issue/PR. I'd suggest we merge this in and open a separate issue to continue the discussion from previous meetings regarding the most robust way to handle SO results.  
Let us know what you think @jenhicks or if you want a separate build to test performance. Thanks

aseth1:
@aymanhab I am unable to associate markers or GRFs with the model. I see one marker at the origin and it does not move. Similarly, if preview experimental data and load GRFS, then I see also see the GRF vertical at the origin as well.
image

It appears the mechanism by which motions are associated is now broken.

tkuchida:
> It appears the mechanism by which motions are associated is now broken.

Possibly related to #612.

aymanhab:
@aseth1 Thanks for continuing to test the PR. I fixed both the issue with IK result containing all states and the Preview/Associate motion data breakage this morning before I checked the boxes on top. Please pull from the branch and let me know if you continue to have issues.

aseth1:
> Please pull from the branch and let me know if you continue to have issues.

Ahh yes, I should have done that! Will try again.

aseth1:
@aymanhab associating and/or loading experimental data animated well. I didn't notice any choppiness or slowdown :+1:.  I am still seeing that the view is zoomed onto the floor for marker preview and whenever the GRF is associated I am seeing the same as @tkuchida in #612 where the view is lost and only hitting one of the projections brings it back. I think that was happening before this fix, so not a new issue, but just a reminder that that aspect is not fixed yet.

aseth1:
Merging this so it can be in a release candidate for others to verify as well.

aymanhab:
Great, thanks much @aseth1

jenhicks:
On my Windows 10 VM, I'm still seeing the same choppy playback for IK, SO (while running the tool and when playing back results). For CMC, I think it solves slowly enough that it animates OK when running ... playback is still choppy.

I was testing with the 3-21 build Ayman posted on dropbox.

PR #611, opened 2018-03-21
Refine windows installer

aymanhab:
This PR incorporates into the ant build script copying the NSIS script and required image file/bitmap for making installer on Windows.
Remaining steps that are not yet automated are: changing Application icon and running NSIS on the script.


Gui/opensim/installer_files/Windows/make_installer.nsi
```diff
@@ -11,7 +11,10 @@
 
   ;Name and file
   Name "OpenSim 4.0 Beta"
-  OutFile "opensim-4.0-win64.exe"
+  OutFile "opensim-4.0Beta-win64.exe"
```
chrisdembia: For Mac, we are able to use Ant to "configure" a similar file to avoid explicitly listing the version number.

aymanhab: Definitely,  where does it get set?

chrisdembia: https://github.com/opensim-org/opensim-gui/blob/587ee9d90b052d1335f99b41be2455717c6caf90/Gui/opensim/build.xml#L236-L241

https://github.com/opensim-org/opensim-gui/blob/587ee9d90b052d1335f99b41be2455717c6caf90/Installer/macOS/Distribution.xml#L5

aymanhab: Yes I saw that but wasn't sure where app.version is set? Now I see it under the Project Properties. Thanks @chrisdembia

Gui/opensim/installer_files/Windows/make_installer.nsi
```diff
@@ -77,7 +77,7 @@ Section "OpenSim Application" SecMain
   WriteRegStr HKCU "Software\OpenSim4.0Beta" "" $INSTDIR
```
chrisdembia: What about the `4.0Beta` in this line?

aymanhab: Oops, thanks for the catch!

aymanhab:
This has no effect on ci since it doesn't actually try to build the installer (it would timeout anyways if it tried).

chrisdembia:
It's a little concerning that copying files is taking so long on Travis.

aymanhab:
Thanks @chrisdembia :+1

PR #614, opened 2018-03-22
fix Analysis Edit Crash

aymanhab:
Fix crash canceling out of Analysis edit that was due to Object slicing (due to safeDownCast). Analysis Edits on Mac and windows can be done and cancelled safely


aymanhab:
Build failures are unrelated, happen after successful build while copying large files for packaging.

PR #615, opened 2018-03-22
use OpenSim instead of opensim for default install dir on windows

aymanhab:
Per feedback from users


aymanhab:
This PR also tricks ci into building a GUI package based on latest opensim-core

PR #616, opened 2018-03-22
[travis] travis_wait for make PrepareInstaller

chrisdembia:
This PR adds `travis_wait` to a line that may take longer than 10 minutes (running pkgbuild). This is just a workaround; ideally, we would find out what's causing `pkgbuild` to take so long.


aymanhab:
I'll close this PR but leave the branch behind in case needed.

Issue #617, opened 2018-03-23
leg6dof9musc replay of IK solution has disconnected patella


aseth1:
### Steps to reproduce
 - load leg6dof9musc.osim model from Models with distribution. 
 - load the Swing/leg69_IK_swing.mot motion file
 - press play.

### Expected result
Swing motion plays and all patella is constrained to the knee. 

### Actual result
Overall leg motion is correct but the patella does not track the femur.
<img width="477" alt="screen shot 2018-03-22 at 9 45 27 pm" src="https://user-images.githubusercontent.com/6494122/37812664-59bd31ea-2e1e-11e8-848a-0a26e1d3d95c.png">

This seems to be a constraint enforcement issue, because when a coordinate slider is manually adjusted, the patella snaps back into place. 

This is likely related to #608 since the vectorized setting of state values does not enforce constraints. It expects that state values were obtained by satisfying constraints. Although IK computes the constrained  `knee_angle_pat_r` value (constrained to be the knee_angle), it does not report the constrained coordinate value, and `Model::formStateStorage` will use its default value.

The fix here is to add a final `Model::assemble(s)` for good measure OR detect when the input files are motions (e.g. does not have a complete state) and then do the assemble.

### Environment and GUI version
Windows [build 0321](https://www.dropbox.com/home/DevBuilds/build0321)

aymanhab:
Thanks for the find @aseth1 Can you advise as to where to assemble()? e.g. right after calling setStateVariableValues? Should we do this only if there are constraints in the model, for efficiency? Also wouldn't that break #540 again if we assemble since constraints will be enforced? Thanks

aseth1:
`Model::assemble()` will only assemble if there are constraints. For #540 the fix was to move clamping within the enforce constraints check, so the assemble should not push the values within the bound. What did you observe in #620?

aymanhab:
Fixed by PR #620, Please verify and close as needed.

jenhicks:
This is fixed on my Mac (3-26 build)

jimmyDunne:
Fixed on W10.

Build 04-27-18

Issue #540, opened 2018-02-26
Motion playback enforces coordinate range (clamped) not enforced during forward simulation


aseth1:
### Steps to reproduce
1.  Load ..Models/BouncingBlock/bouncing_block.osim
2. Simulate for 1s. 
3. Playback motion

### Expected result
Contact sphere (foot) does not penetrate the floor on playback and joint angles are identical to the forward simulation.

### Actual result
<img width="1251" alt="screen shot 2018-02-25 at 6 58 01 pm" src="https://user-images.githubusercontent.com/6494122/36651610-924037d4-1a5e-11e8-8a4f-a6da027608e0.png">

Note that the foot goes through the floor because pin2 does not flex enough. Instead of reaching -131 degrees the playback restricts the coordinate to -114.5 degrees (-2 rad). This is not an issue with the assembly.

tkuchida:
Possibly related to #534.

aymanhab:
@aseth1 Loading the resulting states file from running FD tool and saving results to a file produced identical curve to the one shown above, Is there a better way to compare? Also running the FD tool as described produces the following message in the shell:

AssemblySolver::track() attempt Failed: SimTK Exception thrown at Assembler.cpp:973:
  Method Assembler::track() failed because:
Unable to achieve required assembly error tolerance.
Assembly error tolerance achieved: 3.9279356989219139e-09 required: 1e-10.
Model unable to assemble: AssemblySolver::assemble() Failed: SimTK Exception thrown at Assembler.cpp:897:
  Method Assembler::assemble() failed because:
Unable to achieve required assembly error tolerance.
Assembly error tolerance achieved: 4.5798533854224388e-09 required: 1e-10.

aymanhab:
The GUI calls setStateVariableValue generically on all states, when the call is made with a Coordinate that is clamped, the clamping is enforced even if enforceConstraints is set to false. Would that explain? @aseth1 @tkuchida @chrisdembia Should we be using a different call that sets the Y vector directly? We know these are states at this point.

aymanhab:
https://github.com/opensim-org/opensim-gui/blob/master/Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java#L793

aseth1:
[...]

PR #620, opened 2018-03-23
Call assemble after setStateVariableValues to satisfy constraints

aymanhab:
Address issue #617

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
+++ b/Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
@@ -811,6 +811,8 @@ public class MotionDisplayer {
           // FIX40 speed this up by using map or YIndex
           context.getCurrentStateRef().setTime(assocTime);
           model.setStateVariableValues(context.getCurrentStateRef(), states.getAsVector());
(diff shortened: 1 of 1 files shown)


aymanhab:
Looking for feedback on:
- If states file should contain constrained coordinates (so other code using this e.g. in tools doesn't ned to assemble)
- If we call assemble only in cases where the model contains constraints
- If this interacts with the fix to issue #540

aseth1:
@aymanhab  are these changes in latest sourceforge GUI download?

aseth1:
AFAICT #617 was not fixed in OpenSim 85d9741-2018-03-22. I still see the patella in space.

aymanhab:
No, they're just on a branch PR, if they look ok to you I can merge them.

aymanhab:
Local build with the fix shows patella in right place as expected.
Retested model and motion from 540 and they were still displaying correctly. 
Waiting for review or suggested optimizations if any

aseth1:
I believe the fix will work and will not break #540.

aseth1:
@aymanhab yes, if you just verified, then go ahead and merge.

aymanhab:
Thanks @aseth1 we should have the build from ci shortly, or you can pull the GUI master

PR #632, opened 2018-03-27
latest commit from models repo

aymanhab:
latest models commit


aymanhab:
@jenhicks this should make sure all models are up to date and above ground in next distro. Thanks for reporting.

PR #634, opened 2018-03-27
Update SWIG bindings per API change 032718

aymanhab:
Update bindings to latest API on core

### DIFF
--- a/Gui/opensim/modeling/src/org/opensim/modeling/ExternalForce.java
+++ b/Gui/opensim/modeling/src/org/opensim/modeling/ExternalForce.java
@@ -149,12 +149,8 @@ public class ExternalForce extends Force {
     return opensimSimulationJNI.ExternalForce_append_point_expressed_in_body(swigCPtr, this, value);
   }
 
-  public void constructProperty_point_expressed_in_body() {
-    opensimSimulationJNI.ExternalForce_constructProperty_point_expressed_in_body__SWIG_0(swigCPtr, this);
-  }
-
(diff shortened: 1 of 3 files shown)


aymanhab:
@chrisdembia Is there a reason the constructProperty_xx macros produce public methods? Are these usable from scripting? Any fundamental reason on the core side that these are not protected or private (so they don't pollute the interface for scripting)? Thanks

chrisdembia:
> Is there a reason the constructProperty_xx macros produce public methods? 

Not really, but it could be helpful when/if we allow creating components in scripting.

> Any fundamental reason on the core side that these are not protected or private (so they don't pollute the interface for scripting)?

I don't think these macros can safely use the visibility keywords, because the macro doesn't know what section it's in.

aymanhab:
Thanks @chrisdembia for the clarification. While the macro can't tell the context, the class writer/header can and should handle this properly so as not to pollute the public interface and methods searched in both Doxygen and by Scripting. I don't think this is needed even for extension from Scripting which would use the swig/director and keep its own internals rather than use the Maco-centric Property system.

chrisdembia:
> the class writer/header can and should handle this properly so as not to pollute the public interface and methods searched in both Doxygen and by Scripting.

The macro contains functions that *should* be public, though.

aymanhab:
Thanks for the feedback @chrisdembia I opened issue #2135 [https://github.com/opensim-org/opensim-core/issues/2135] to track.

Issue #612, opened 2018-03-21
Visualizer Window goes blank after Associate Motion Data operation


tkuchida:
### Steps to reproduce
- open Models\Gait10dof18musc\gait10dof18musc.osim
- File > Load Motion…
- select Models\Gait10dof18musc\OutputReference\IK\subject01_walk_IK.mot
- in Navigator, right-click *Coordinates* and select *Associate Motion Data…*
- select Models\Gait10dof18musc\ExperimentalData\subject01_walk_grf.mot

### Expected result
GRF vectors appear in Visualizer Window.

### Actual result
Visualizer Window goes blank. Clicking/dragging in Visualizer Window has no effect. Model is rendered again upon clicking the button named "Left".

![](https://user-images.githubusercontent.com/4203505/37742442-d242af70-2d22-11e8-921c-66354f16aaf4.png)

### Environment and GUI version
AppVeyor artifact OpenSim-587ee9d9-2018-03-21.zip

jenhicks:
Fixed for me on Mac, following the steps that Tom outlined.

jimmyDunne:
Fixed on W10, followed instructions above.

System:W10
Build:04-27-2018

PR #640, opened 2018-03-30
Avoid sending uninitialized bounding box to visualizer 

aymanhab:
when loading experimental forces, don't send uninitialized bounding box to visualizer


aymanhab:
This fixes issue #612

PR #645, opened 2018-04-03
Fix model offset and associate motion on sync.

aymanhab:
Display offset dialog fixed to set offset rather than position and associated motions are retained on sync.


aymanhab:
Per discussion with @chrisdembia the changes to version numbers are only in resource files rather than the app.version used to name the result package.

Issue #522, opened 2018-02-15
Show/hide errors when model is not the current model (ToyLandingModel)


nickbianco:
While running the ToyLandingModel tutorial, I found that showing and hiding the model behaved inconsistently for models not designated as the current model. (Windows, AppVeyor artifact OpenSim-dd7dc47b-2018-02-11).

To reproduce:
- Open the ToyLandingModel.osim and ToyLandingModel_AFO.osim models
- Set the ToyLandingModel.osim model as the current model
- Right click the ToyLandingModel_AFO.osim model in the Navigator and select Display --> Hide. The footplate geometry _will not_ be hidden. Trying to hide the r_AFO_footplate body directly will not hide the geometry either.
- Set the ToyLandingModel_AFO.osim as the current model. You can now hide the footplate geometry, either by hiding the entire model, or by hiding through the r_AFO_footplate body directly.
- Keep ToyLandingModel_AFO.osim hidden and set ToyLandingModel.osim as the current model.
- Now try to show ToyLandingModel_AFO.osim (right click --> Display --> Show). Depending on whether or not you previously hid the entire model or only hid the footplate to remove the footplate geometry, different errors will occur. The model will either not show at all (if you previously hid the entire model), or the model will show but without the footplate geometry (if you only hid the footplate).

jenhicks:
Show/hide seems to be working as expected now on my Mac.

Issue #644, opened 2018-04-02
Unhiding a hidden model reveals all its wrapping surfaces and turns muscles red


aseth1:
### Steps to reproduce
- Load two models.
- Load two motions, one for each model.
- Associate motion data.
- hide one Model
- hide the other
- unhide a the first model

### Expected result
The unhidden model appears exactly as it appeared when it was hidden.

### Actual result
- Wrapping objects now appear
- muscles turn red irrespective of the state they were in when hidden

<img width="940" alt="screen shot 2018-04-01 at 7 25 38 pm" src="https://user-images.githubusercontent.com/6494122/38180653-115ee3d6-35e3-11e8-941e-3250c8829dbe.png">

### Environment and GUI version
Mac OpenSim 651d29f-2018-03-30

jenhicks:
I tested this with Apoorva's model on my Mac and I don't see the wrapping surfaces appear or muscles turn red after following the steps that @aseth1 describes.

jimmyDunne:
Used Apoorva's model and states files. The wrapping surfaces do not open when 'show' is selected on the top level model. 

System:W10
Build:04-27-2018

PR #649, opened 2018-04-05
Separate command for model Show/Hide similar to ShowOnly 

aymanhab:
These commands now operate on the Model visualization as a whole rather than descend to leaf nodes so they leave user set visualization preferences intact.


aymanhab:
Fixes issues #522 #644

Issue #654, opened 2018-04-09
An extra marker is shown when experimental data is previewed the first time after starting the GUI


tkuchida:
### Steps to reproduce
1. Start OpenSim
2. File -> Preview Experimental Data...
3. Select Gait10dof18musc\ExperimentalData\subject01_walk.trc
4. File -> Close All
5. File -> Preview Experimental Data...
6. Select Gait10dof18musc\ExperimentalData\subject01_walk.trc

### Expected result
- Data loads, visualizer displays first frame, camera adjusts appropriately
- The user should see no difference in the GUI following steps 3 and 6

### Actual result
**After step 3 (red circle added):**
![](https://user-images.githubusercontent.com/4203505/38528119-0cd69a26-3c13-11e8-9c8a-29613892dda7.png)

**After step 6:**
![](https://user-images.githubusercontent.com/4203505/38528052-cfb0a0e2-3c12-11e8-83c2-42ef560e4b00.png)

### Environment and GUI version
AppVeyor artifact OpenSim-587ee9d9-2018-03-21.zip

jimmyDunne:
Repeating the steps above, with described data as well as new files, the 'Ghost' marker did not appear.

tkuchida:
Fixed on AppVeyor artifact OpenSim-d8bfa962-2018-05-09-win64.exe

Issue #653, opened 2018-04-09
Visualizer doesn't display first frame when previewing experimental data


tkuchida:
### Steps to reproduce
- File -> Preview Experimental Data...
- Select Gait10dof18musc\ExperimentalData\subject01_walk.trc

### Expected result
- Data loads, visualizer displays first frame, camera adjusts appropriately

### Actual result
Camera is zoomed in on one marker (reported in #508):
![](https://user-images.githubusercontent.com/4203505/38527683-decf3a68-3c10-11e8-81fe-daac4676ddc3.png)

Unlike in #508, there are now no other markers after zooming out:
![](https://user-images.githubusercontent.com/4203505/38527689-e37dc25a-3c10-11e8-8378-e6450c00ca5c.png)

Markers appear after forcing time update (in this case, I clicked the :previous_track_button: button):
![](https://user-images.githubusercontent.com/4203505/38527694-e99d1e24-3c10-11e8-90f1-24ddf0ff19f4.png)

### Environment and GUI version
AppVeyor artifact OpenSim-587ee9d9-2018-03-21.zip

jimmyDunne:
Markers loaded and (appears) to open at correct frame of data (no erroneous marker on first frame)

tkuchida:
@jimmyDunne What environment and build were you using? Please test with this file: [test.zip](https://github.com/opensim-org/opensim-gui/files/1931194/test.zip) (contains test.trc). On my machine using AppVeyor artifact OpenSim-dbbbb0c5-2018-04-16.zip, when I preview the first time, the camera is looking at nothing and I can't find the markers:
![](https://user-images.githubusercontent.com/4203505/39037086-511f08a2-4434-11e8-94c2-bea050b3686a.png)
The markers appear after *File > Close All* and previewing a second time.

jimmyDunne:
Hey, @tkuchida. I used the .trc you included (thanks!).  Below is a gif of me trying to hunt for the markers. When I 'preview experimental data', there are no markers seen. I then have to search the space by trying a few different views until I 'spot' a marker and focus in. 

![marker_hunt](https://user-images.githubusercontent.com/5296507/39062464-6f4723da-447c-11e8-9989-cbf708c54910.gif)

[...]

Issue #471, opened 2018-01-11
Experimental markers cannot be hidden


tkuchida:
Tested on leg6dof9musc.osim. Unable to hide all markers (right-click Markers in the Navigator window and select Display > Hide) and unable to hide individual marker (right-click a marker name in the Navigator window and select Display > Hide). No messages in the Message window and no red-circle errors. Marker does disappear if deleted (right-click a marker name in the Navigator window and select Delete).

aymanhab:
@aseth1 will look into how markers are generating decorations

aymanhab:
For the record,  in 3.3 the markers could be hidden in GUI but that was not persistent since there was no xml artifact. Adding Appearance to marker will allow the display to be persistent, and may allow changing the color, size of markers (functionality that didn't exist pre 4.0).

aymanhab:
If the conclusion is to make non-persistent GUI changes then please reassign to me.

aseth1:
Try to expose the Sphere geometry Appearance directly.

aseth1:
YEs, OK for hidden to not be persistent.

tkuchida:
Fixed in AppVeyor artifact OpenSim-dd7dc47b-2018-02-11.zip (tested Show/Show Only/Hide). @aseth1: Please close if verified (or if not necessary to verify) on :apple:.

aseth1:
In Mac OpenSim 651d29f-2018-03-30, I am unable to hide a subset of experimental markers.

aymanhab:
@aseth1 Experimental markers are totally different creatures since they are not Components that belong to the model. Please open a separate issue.

jenhicks:
I cannot hide experimental markers. Should I move this back to next up @aymanhab ? Testing on the latest Mac build.

jenhicks:
Ooops .. realized this isn't in my build yet.

jimmyDunne:
Experimental markers can be hidden, shown, and shown only. 

![opensim_markers.gif](https://images.zenhubusercontent.com/544fc8297c6ccc6277a5b651/0ad5a4b3-0061-424b-85ff-b1fb2a13fe47)

Windows 10 
opensim-04-11-2018

Issue #397, opened 2017-09-29
Items under "Display" in context menu have no effect for experimental data


tkuchida:
### Steps to reproduce
- open Models\Gait10dof18musc\subject01.osim from 3.3 distribution
- File -> Preview Experimental Data…, select Models\Gait10dof18musc\ExperimentalData\subject01_walk.trc
- right-click ExperimentalData_0 in Navigator panel; Display->Show, etc. have no effect

opensim4.0PreviewWin64_0929.zip

jimmyDunne:
For experimental data open in the navigator, the context menu for 'hide' and 'show' do not do anything. 'show only' works, and will make other models or data disappear. (The video capture method I am using doesn't show the context menus)

If I hide any markers individually, then go to the top level and select 'show', the hidden ones don't reappear. If, on the top level, you select 'show' or 'hide', it should have that effect on all markers. 

![markers_hide](https://user-images.githubusercontent.com/5296507/38648549-51fb3ed4-3de1-11e8-8f60-7b25bf907ed6.gif)

Windows 10
opensim-04-11-2018

aymanhab:
showOnly on a model will hide other models
operations on top level model used to visualize data operate on the whole model without changing the user preferences/choices underneath it (that was the bug in issue #522 and #644)
Somewhere else there was a report that the functionality works for hide/show/showOnly. 😕 
Behavior is essentially the same as in 3.3 unless there's a specific unexpected behavior

PR #658, opened 2018-04-11
Use first frame of data for visualization

aymanhab:
Rearrange initialization of MotionDisplayer, use data from first frame to create visuals. Also implement the Display menu under ExperimentalData objects for Hide/Show/ShowOnly remove trails


aymanhab:
Address issues #397 #471 #653 #654

Issue #352, opened 2017-08-03
Rendering artifacts in wrap geometry


tkuchida:
![geometry](https://user-images.githubusercontent.com/4203505/28940444-9b68fb8c-7849-11e7-9fbe-36364d8cfd55.png)

arm26.osim; 4.0 alpha, ISB release; Win 8.1

jenhicks:
@tkuchida Can you check whether this is still an issue for current default visualization settings?

tkuchida:
Still an issue in AppVeyor artifact OpenSim-be3c7698-2018-04-05.zip.

**Camera angle 1**
![](https://user-images.githubusercontent.com/4203505/38397427-649ba3b0-38f2-11e8-915a-e55c4f22ef35.png)

**Camera angle 2**
![](https://user-images.githubusercontent.com/4203505/38397407-58f22412-38f2-11e8-9294-ab59bee716ea.png)

aymanhab:
@tkuchida kinks around wrapping have always been smoothed out by introducing a number of intermediate points on the opensim-core side rather than by changing the path in visualization.

tkuchida:
> kinks around wrapping have always been smoothed out by introducing a number of intermediate points on the opensim-core side rather than by changing the path in visualization

This Issue has to do with rendering artifacts in the wrap geometry (the cyan shapes shown in the images above); perhaps your comment was intended for #765?

tkuchida:
See https://stackoverflow.com/questions/13221328/artifacts-when-rendering-both-sides-of-a-transparent-object-with-three-js

tkuchida:
Another test case (arm26.osim): floor is visible through wrap geometry but bone should be rendered there instead:

![](https://user-images.githubusercontent.com/4203505/40275670-940ad838-5bab-11e8-983f-c8a14ea3b637.png)

tkuchida:
I see fewer artifacts if only front faces are rendered (see [Material.side](https://threejs.org/docs/#api/materials/Material.side)):

| **THREE.DoubleSide (current setting)** | **THREE.FrontSide** |
| --- | --- |
| ![](https://user-images.githubusercontent.com/4203505/41026836-680b8fe8-692a-11e8-8f50-5b5d42453df2.png) | ![](https://user-images.githubusercontent.com/4203505/41026844-6bf32abc-692a-11e8-8898-cb9c3bedad9f.png) |

[...]

PR #659, opened 2018-04-11
Edit analytic geometry shapes

aymanhab:
Replace visualization of analytic geometric types (Sphere, Cylinders, etc.) when parameters change since parameters can't be changed post hoc on the visualizer side (other than scale factors)


aymanhab:
Address issue #436, and partly #352 by refining resolution of all built in Spheres, Cylinders, etc.

Issue #539, opened 2018-02-26
Muscle Display Radius has no effect on muscle display in GUI


aseth1:
### Steps to reproduce
Edit->Preferences...->Muscle Display Radius changed to 0.001 (from 0.05).

### Expected result
Expected muscles to appear thinner.

### Actual result
No effect.
(include screen grabs where appropriate)

### Environment and GUI version
Mac build: OpenSim c596914-2018-02-24

bspineda:
Here is an image of the rat model. This will need to be fixed. 
![rat](https://user-images.githubusercontent.com/32551585/36701415-2cfde84e-1b08-11e8-8e37-033ea5ce9c7b.png)

jimmyDunne:
Changing the Muscle Display radius does not change a models muscle radius. You have to close and re-open the model for the altered Muscle Display Radius to take effect. 

There are also two entries for Muscle Display Radius in the Preferences Pane. 

![muscle_radius_preferences](https://user-images.githubusercontent.com/5296507/39588705-28b042fa-4eb1-11e8-800c-86e36b9889be.png)

W10
OpenSim-ba958aac-2018-04-30_threejs_r91.zip

aymanhab:
Some preferences take effect immediately and some upon loading a new model, considering that they persist I wouldn't consider this a bug. The duplicate Preference is a bug however that I will investigate. Thanks for testing @jimmyDunne

aymanhab:
Actually the names are different by a typo "Display" vs. "Dsiplay"! fixed to unify and retired the misspelled one!

jimmyDunne:
Based on Ajay's post, the desire would be for the muscles to change radius immediately- I would like to see the same behaviour. If a user is trying to make a figure and has to have to iterate by closing down the whole application, I don't think that is great.

aymanhab:
@jimmyDunne Recreating the geometry for all muscles on the fly to change radius, while possible, seems like an overkill of a functionality to provide considering that users can load another instance of the model without closing the app to get the same effect. 
While creating a figure is very important, the more likely use case is either you always prefer the muscles to be thinner/thicker or you're using non-human model where this change is rather long term.

jimmyDunne:
Thanks for the good perspective, Ayman. @aseth1, are you okay with me moving this to 'done'? The muscle display radius does change, just after a reboot. I will add a documentation tag so that this behaviour is known.

aseth1:
[...]

PR #692, opened 2018-04-27
Pass preference to visualizer to set radius of skinnedMuscle

aymanhab:
Communicate user preference for muscle display radius to visualizer


aymanhab:
Fix issue #539, GUI side.

Issue #676, opened 2018-04-20
Global Frame Visualization Button: Label X/Y/Z or Add Tooltip


jimmyDunne:
### Steps to reproduce
Try to view the Ground Frame (Global origin)

### Expected result
I should have a button to draw a Global Frame on the screen that gives me (0,0,0) and orientation

![global_button_33](https://user-images.githubusercontent.com/5296507/39063214-17a555ea-447f-11e8-9a9d-a38790096dfc.png)

### Actual result
There is no button to draw a Global Frame on the screem

![global_button_40](https://user-images.githubusercontent.com/5296507/39063224-239efee6-447f-11e8-8399-d64563e628c0.png)

### Environment and GUI version
W10
04-16-2018

jimmyDunne:
Frame Button exists and you can toggle on and off the global frame. 🍾 🎆

tkuchida:
Button exists and works in AppVeyor artifact OpenSim-bf3e61b9-2018-05-09-win64.exe, but "X", "Y" and "Z" labels are not shown. Ideally these labels would be added to the scene as in 3.3; otherwise, perhaps define the mapping between color and axis in the button's tooltip (which currently does not exist). Perhaps something like the following:
<pre>
Show/hide global axes (X/Y/Z = red/green/blue)
</pre>

aymanhab:
Labels, tooltips are Global frame X(Red) Y(Green) Z(Blue).

aseth1:
I verified on Windows OpenSim_cea85df3-2018-07-18 that -X, X, -Y, Y, -Z, Z Tool tips appear and that the global axes colors are described in the tool of the toggle for global axes display.

jenhicks:
Thanks ... moving to verified on windows + mac

PR #694, opened 2018-04-27
Include fix to muscle display radius and show/toggle global frame 

aymanhab:
Tis just upgrades the module to latest on threejs master


aymanhab:
Fix issue #539 #676

PR #699, opened 2018-05-02
Create PULL_REQUEST_TEMPLATE.md

jimmyDunne:
Adds descriptions to PRs that will help testers with verifying changes


aymanhab:
Good idea, the problem with this repo though is that much is happening on submodules that may not be shown here in details e.g. changes to threejs, or visualizer or core changes but still helpful. Thanks @chrisdembia @jimmyDunne

PR #708, opened 2018-05-03
Fix typo in "Muscle Display Radius" preference 

aymanhab:
This caused Preference to appear twice, also this PR retires some prefs that were used during development

Fixes issue #539 by removing duplicate preference reported earlier

### Brief summary of changes
Fix typo and retire some Preferences that were introduced during development

### Testing I've completed
Launched application, opened preference dialog, no repeated Preferences

### CHANGELOG.md (choose one)

- We should update current list of Preferences and when they do take effect in user guide.

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/actions/EditPreferencesJPanel.java
+++ b/Gui/opensim/view/src/org/opensim/view/actions/EditPreferencesJPanel.java
@@ -70,8 +70,11 @@ public class EditPreferencesJPanel extends javax.swing.JPanel {
        "AntiAliasingFrames",
        "Marker Display Radius",
        "NonCurrentModelOpacity",
-       "Refresh Rate (ms.)"
-       
+       "Refresh Rate (ms.)",
+       "Muscle Dsiplay Radius",
+       "BuildDate",
+       "One Material Meshes",
+       "Save Movie Frames"
    };
    /** Creates new form EditPreferencesJPanel */
    public EditPreferencesJPanel() throws BackingStoreException {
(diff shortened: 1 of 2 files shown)


Gui/opensim/view/src/org/opensim/view/actions/EditPreferencesJPanel.java
```diff
@@ -70,8 +70,11 @@
        "AntiAliasingFrames",
        "Marker Display Radius",
        "NonCurrentModelOpacity",
-       "Refresh Rate (ms.)"
-       
+       "Refresh Rate (ms.)",
+       "Muscle Dsiplay Radius",
```
tkuchida: Dsiplay -> Display ?

aymanhab: @tkuchida I'm retiring the name with typo so users don't see it anymore, the correctly spelled "Muscle Display Radius" will be kept. Thanks for the review :+1:

tkuchida: Ok, sounds good. I must be missing something here: GitHub is indicating that you're *adding* "Muscle Dsiplay Radius" which contains the typo.

aymanhab: No, you're absolutely right but I'm adding it to the deprecated/hidden list so users who got it by mistake in a previous build don't see it anymore.

tkuchida: Ah, ok sorry to hold up the train :steam_locomotive:; I didn't realize there was a list of all the old property names.

aymanhab: Unfortunately we need to keep these since registry entries on Windows stay forever! Thanks again for the review 👍

Issue #553, opened 2018-02-26
Trying to run FD on WalkerModelAddMagnet_Feet hangs


aymanhab:
### Steps to reproduce
Load, run FD from tool or running dude

### Expected result
Runs

### Actual result
Nothing happens
(include screen grabs where appropriate)

aymanhab:
[WalkerModelAddMagnet_Feet.txt](https://github.com/opensim-org/opensim-gui/files/1760380/WalkerModelAddMagnet_Feet.txt)

aymanhab:
Encountered by Scott during feedback session

tkuchida:
Unable to reproduce:
![](https://user-images.githubusercontent.com/4203505/36698130-e66d83ee-1afd-11e8-8fc3-2e211c8f725b.png)
What .obj file should I be using?

aymanhab:
Some context? Where the model came from? Does it run in 3.3? Where did it get the files then?

aymanhab:
Sorry @tkuchida I didn't realize that I opened this issue 😕

tkuchida:
I renamed the file (attached above) from .txt to .osim, then replaced all instances of "ThinHalfCylinder100mmby50mm.obj" with "blockMesh.obj" (using blockMesh.obj from the Geometry directory). Model opens, but :one: is not centered in the Visualizer window (so perhaps a malfunction occurred silently---but no red circle error); and :two: the :running_man: button doesn't do anything. Forward Tool also doesn't do anything after clicking *Run*.

aymanhab:
Can you upload the modified .osim (renamed to .txt) so I can try to run in debug mode? Thanks @tkuchida

tkuchida:
@aymanhab [WalkerModelAddMagnet_Feet_modified.txt](https://github.com/opensim-org/opensim-gui/files/1760567/WalkerModelAddMagnet_Feet_modified.txt)

aymanhab:
Inertias all messed up: Messages from debug build below, not sure if only reason yet

WARNING: Body Pelvis has invalid inertia. 
SimTK Exception thrown at MassProperties.h:548:
  Error detected by Simbody method Inertia::Inertia(moments,products): Diagonals of an Inertia matrix must satisfy the triangle inequality; got 0.0001,0.00666667,0.0001.
[...]

PR #724, opened 2018-05-09
Fix path maintained by internal models used to run tools

aymanhab:
Fixes issue #553

### Brief summary of changes
Removed calls to setInputFilename("") on throwaway models used to run tools in GUI

### Testing I've completed
Successful run of FD on model with contact meshes, launched other tools on the same model.

### CHANGELOG.md (choose one)

- no need to update because it is a bugfix


aymanhab:
@jimmyDunne I only tested on the model in bug report and didn't have data files to run all tools. If you can try a gait model with contact/AFO and exercise the variety of tools that would be great.

Issue #712, opened 2018-05-04
NSIS Uninstaller.exe does not remove the installed files.


chrisdembia:
### Steps to reproduce

Install OpenSim using the NSIS installer. Then go into the installation and run the Uninstall.exe.

### Expected result

The OpenSim files are removed.

### Actual result

Only Uninstall.exe is removed.

### Environment and GUI version

Master branch, building the installer locally.

jimmyDunne:
Uninstaller deletes folder and files and removes the path from system path. 
W10, OpenSim-32ef4769-2018-05-12-win64.exe

PR #732, opened 2018-05-10
Allow uninstaller to remove non-empty install directory

aymanhab:
Fixes issue #712

### Brief summary of changes
Uninstaller on Windows now removes the installation folder even if non-empty, added switch

### Testing I've completed
Built installer and tried install/uninstall on my local machine. All worked as expected

### CHANGELOG.md (choose one)


aymanhab:
ci doesn't exercise the installer, no need to wait for it.

Issue #675, opened 2018-04-20
Camera Changes when data is opened 


jimmyDunne:
### Steps to reproduce
Open experimental marker data

### Expected result
The camera stays where it is.

### Actual result
Camera zooms down onto the ground origin.

![camera_zoom](https://user-images.githubusercontent.com/5296507/39063277-65808d52-447f-11e8-9a08-3f7029d937bf.gif)

### Environment and GUI version
W10
04-16-2018

Issue #698, opened 2018-05-02
I can delete the Sun!


jimmyDunne:
### Steps to reproduce
Open model, open scene ligh, expand floor dropdown, click in 'Height' text box, hit backspace on keyboard.

### Expected result
I can delete the numbers in the text box

### Actual result
I get the rush of a Supervillian by deleting the Sun! 

![sunlight](https://user-images.githubusercontent.com/5296507/39537615-fcb09000-4dee-11e8-90ef-2fdd841a9667.gif)

### Environment and GUI version
W10
opensim-04-30-2018

aymanhab:
Actually you can't, you can just move it underground! Hope that keeps your Supervillian spirits up

aymanhab:
Fixed by PR in threejs repo. Will appear in next threejs upgrade

jimmyDunne:
I'm not sure how to test this. It looks like the ability to view the light and move it is now gone! Was this intentional? Now, under the lighting menu, there are only text boxes that do something (that is unclear) to the lighting.

aymanhab:
@jimmyDunne this was part of the plan to make the placement of light more reproducible so you see numbers for x,y,z of scene light. It doesn't help that you can't see the light source itself 😃 I'm working on a compromise and appreciate your input, so drop by when you have a chance to discuss.

tkuchida:
I looked at this as well. The shadow doesn't change when the X/Y/Z fields are edited. I think we need to be able to move the light that casts the shadow in case the shadow gets cut off (#477):

![](https://user-images.githubusercontent.com/4203505/40077032-4fb9c42e-5835-11e8-936f-4caf14d28589.png)
75fdf108-2018-05-11

aseth1:
Following-up on @tkuchida since this issue remains in **Needs Verifying**.
The lighting and shadows are inconsistent with the shading of display geometry.
![image](https://user-images.githubusercontent.com/6494122/40325876-2889e034-5cf2-11e8-86c3-db33fb9170c2.png)

Furthermore, the lighting is not persistent and if I tumble the view in the slightest it returns to what it was previously but none of the lighting values change!
![image](https://user-images.githubusercontent.com/6494122/40326013-a44026d4-5cf2-11e8-830e-a50c78efcbde.png)

More disappointing, however, is the inability to graphically move the light source in the 3D view and see the shadows move. Ideally the position values in the text fields would update and the user could store them as presets. 
[...]

PR #733, opened 2018-05-10
Bring over updates from threejs and opensim-visualizer

aymanhab:
Fixes issue #675, #698

### Brief summary of changes
Fixes merged into threejs module are brought over.

### Testing I've completed
Followed steps in issues and got expected behavior

### CHANGELOG.md (choose one)

- no need to update because bug fixes


aseth1:
This "fix" appears to prevent using the delete and other keys in the visualizer window. In that event how does one correct typos/errors when entering a values like floor height, or  other value in the view controls? Are they dictated by another key handler?

aymanhab:
@aseth1 This code has nothing to do with the visualizer controls menus/text-boxes.  The "fix" does exactly what is intended for, which is to disallow users changing the visuals by deleting visuals, or undoing delete etc. behind the GUI's back, that was never intended. Other parts/libraries have their own handlers.

Issue #686, opened 2018-04-26
Unexpected behavior when hiding/showing bodies


tkuchida:
### Steps to reproduce
1. open arm26.osim
2. in Navigator, right-click *ground* and select *Display > Hide*
3. in Navigator, right-click *Bodies* and select *Display > Show*

### Expected result
Geometry attached to bodies appears after operation 3.

### Actual result
Visualizer Window remains blank. The logic here seems inconsistent: right-clicking *ground* and selecting *Display > Hide* causes everything to disappear, so it seems like hiding *ground* propagates down the ownership tree to everything below it. If so, right-clicking *Bodies* and selecting *Display > Show* should make the bodies (and everything below them in the ownership tree) reappear. Furthermore, if only some components are shown and the user right-clicks *ground* and selects *Display > Show*, nothing happens (expected everything to appear).

### Environment and GUI version
AppVeyor artifact OpenSim-e5f97340-2018-04-20.zip

jenhicks:
Same on Mac (5-31 build)

PR #735, opened 2018-05-11
Don't apply hide/show to aggregate Nodes directly, only leaf nodes

aymanhab:
Fixes issue #686 

### Brief summary of changes
When applying hide/show to nodes, apply only to leaf nodes in navigator view

### Testing I've completed
Sequence of hide/show/sjowOnly

### CHANGELOG.md (choose one)

- no need to update because behavior is unchanged from 3.3


aseth1:
@aymanhab I don't see how this is the correct fix to #686 since now it seems that one cannot select hide/show all bodies, joints, muscles by selecting the parent (Set) node. Is that right? Have you verified that behavior is not broken before requiring user testing?

aymanhab:
Yes @aseth1 it is the correct fix, traversal of the nodes still processes all the nodes on the tree but collection of objects works only on leaf nodes. This avoids hiding a "body" or "ground" resulting in the hiding of everything attached to it in a scene graph sense.

Issue #761, opened 2018-05-18
Pink model markers do not appear to be spheres


tkuchida:
### Steps to reproduce
Open a model with markers; load corresponding experimental marker data.

### Expected result
Blue experimental markers and pink model markers intersect in the Visualizer Window as if they are spheres.

### Actual result
The blue experimental markers seem to "poke through" the pink model markers when the markers are close (which is often the case):

![](https://user-images.githubusercontent.com/4203505/40210611-a7c109cc-59fa-11e8-9cc4-54fc1839829a.png)

Not sure how this can happen (perhaps the pink spheres have front face culling on rather than back face?), but it doesn't look so good and might make it difficult for users to place markers on the skeleton in the correct positions, which is a critical step in the pipeline.

### Environment and GUI version
75fdf108-2018-05-11

tkuchida:
Perhaps the [`Material.side`](https://threejs.org/docs/#api/materials/Material.side) has been set to `THREE.BackSide` rather than `THREE.FrontSide`?

jimmyDunne:
Model makers seem to be spheres. I don't see the issue that Tom reported. 

@tkuchida, I will move this to verified- please let me know if you see this behaviour again? Thanks.

W10, OpenSim-2888c1c6-2018-05-28-win64.exe

jenhicks:
Marker intersections appear as I would expect on Mac, as well

PR #767, opened 2018-05-19
Use MeshStandardMaterial for ModelMarker with 2 sidedness 

aymanhab:
This makes model markers in line with other meshes. Emissive color is set to pink as well so the markers are not dark if in shadow and indistinguishable from  Experimental markers

Fixes issue #761 

### Brief summary of changes
Changed values and type of Material pre-created for Model markers

### Testing I've completed
Loaded models with markers 

### CHANGELOG.md (choose one)

- no need to update, bugfix

### DIFF
--- a/Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
+++ b/Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
@@ -1260,8 +1260,11 @@ public class ModelVisualizationJson extends JSONObject {
         UUID mat_uuid = UUID.randomUUID();
         mat_json.put("uuid", mat_uuid.toString());
         mat_json.put("name", "MarkerMat");
-        mat_json.put("type", "MeshPhongMaterial");
-        mat_json.put("shininess", 30);
+        mat_json.put("type", "MeshStandardMaterial");
+        mat_json.put("transparent", true);
+        mat_json.put("metalness", 0);
+        mat_json.put("roughness", 1.0);
+        mat_json.put("side", 2);
         //mat_json.put("transparent", true);
         String colorString = JSONUtilities.mapColorToRGBA(hints.get_marker_color());
         mat_json.put("color", colorString);


tkuchida:
Sorry, but I don't really have a way of reviewing this change (I haven't been able to build the GUI on my machine). However, setting `transparent` to `true` and also `side` to `2` suggests that we might end up with rendering artifacts similar to what we get for wrap objects (see #352 and https://stackoverflow.com/questions/13221328/artifacts-when-rendering-both-sides-of-a-transparent-object-with-three-js). Do you see these artifacts on your local build? Also, the blue experimental markers don't appear to be transparent, and I think the experimental and model markers should match in all respects except for color.

aymanhab:
Tom, this change passes values to the visualizer from GUI. If you do the exact change of the 2-3 properties on marker material in expert mode you'll see the result of the change, that's why I asked for your review. Transparency value and opacity are the same as experimental markers, and can't be changed in GUI but are set to avoid ambiguous default properties.

tkuchida:
I gave up trying to get the pose to match, so I just dragged one of the model markers to overlap with an experimental marker. Here's what it looks like before the changes (using 75fdf108-2018-05-11):


Here's what it looks like after applying the changes in the PR (set material type to MeshStandardMaterial, transparent = true, metalness = 0, roughness = 1, side = double):


Here's what it looks like after setting `color = emissive = RGB(255, 192, 255)` (my guess at what the colors are being set to):


Issue #761 is resolved, but I think the appearance of markers can still be improved in two ways:
1. The experimental markers are not smooth (should use the same number of polygons as model markers).
2. The model marker is too shiny, and will be lost against a white background. Perhaps use `color = emissive = RGB(255, 112, 255)` (see below)?

tkuchida:
> I don't have ready files that have coincident markers

I've been using these:
<pre>
Resources\Models\Gait10dof18musc\OutputReference\subject01.osim
Resources\Models\Gait10dof18musc\OutputReference\IK\subject01_walk_IK.mot
Resources\Models\Gait10dof18musc\OutputReference\ExperimentalData\subject01_walk.trc
</pre>

> The model marker colors are provided by the API (now hard-coded as pink there), so I can't pick arbitrary colors for emissive, now using white because black results in model markers looking black/dark if in shade. Let me know if you have other ideas suggestions

Please provide a screencap or let me know what RGB values get assigned for color. Alternatively, you could merge and we can leave the Issue open to tweak colors on the next cycle.

aymanhab:
Color computed from RGB values provided by API is 255, 153, 204 and is used as both Color and Emissive.

aymanhab:
Thanks for the review @tkuchida Much appreciated. Will merge and we can revisit if there's a general way to set "emissive" value that doesn't washout.

aymanhab:
capturemarkers

Issue #279, opened 2017-07-12
Changing motions from one with muscle states to without: color of muscles is carried over 


jenhicks:
This happened for me in the tutorial when I switched from the SO results to IK results

aymanhab:
This was folded into #283 where loading any motion now, we convert into States using formStateStorage.  Please verify and close or report unexpected behavior. Thanks

jimmyDunne:
I opened a model and two motions; one a states file from CMC, and the other is an RRA sto file from RRA. When I switch from the CMC states to the RRA kinematics, whatever the current muscle colors were during the CMC playback are persistent. 

![opensim 4 0 4_11_2018 2_03_10 pm](https://user-images.githubusercontent.com/5296507/38643571-662b4312-3d92-11e8-96e0-2f194fb0b109.gif)

jenhicks:
When you load coordinates, make full state with muscles to blue. @aymanhab will investigate

aymanhab:
This actually is exactly how Model.formStateStorage() works which is correctly invoked in this case. For unspecifed states (e.g. muscle states) it keeps them at current value. If this is unexpected or a different behavior is expected pleaseopen an issue for opensim-core and when a solution is available at the API level we can wire it from the  GUI. @chrisdembia please let me know if this is different from your expectations. Thanks.

jimmyDunne:
Ayman, it would be great if you could pass on more details on how this feature works. It feels like the rest of the team aren't sure of how this is implemented and it would be awesome to get some further insight. What class/component actually draws the colors? What class/component holds/computes the data? It sounds like Model.formStateStorage() will compute some data that then gets passed to something that draws it on the screen, correct? When we have multiple loaded or motions or states, does Model.formStateStorage() get computed each time you change the current motion or state or does it called once when the data is loaded and stored somewhere for recall when a user changes the current motion or state? It would be really helpful for us to know the design. 

We had discussed that when we switch between a loaded state with muscle activations to a motion with only coordinate values, the muscles should become blue— the recognized color for zero activation. Sorry if that was unclear from the meeting?

If this there is an improvement that should be made to OpenSim-core to help the visualizer, it would be most helpful for you to create and write the issue. You are by far the most familiar with the GUI and the communication between the GUI and the API. I would find it daunting to write an issue on the this that capture all the nuance.

aymanhab:
[...]

PR #804, opened 2018-05-30
Explicitly call resetStateToDefault before loading motions

aymanhab:
This resets states to default values before loading a motion, making sure colors/states don't carry over between motions
Fixes issue #279 

### Brief summary of changes
Added the call to reset states to default before calling Model.formStateStorage
### Testing I've completed
tested motions in issue reported and colors don't carry over.
Looking for feedback on whether there are other side effects to the call.

### CHANGELOG.md (choose one)

- no need to update bugfix


aymanhab:
Can you review or suggest a different solution/workaround @aseth1 ? Thanks.

PR #806, opened 2018-05-30
Run vcredist_x64.exe in NSIS installer.

### LINKED ISSUE #744: Installer should run Visual C++ Redistributable installer
Here are MIcrosoft's guidelines for distributing the Visual C++ runtime: https://docs.microsoft.com/en-us/cpp/ide/deployment-in-visual-cpp

Right now, we expect the user to either have the correct Visual C++ runtimes via "central deployment" or that the user is obtaining the runtimes through the "local deployment" in `opensim-core`.  In the article above, Microsoft recommends using "central deployment," which means running the Visual C++ Redistributable installer as part of our installer. 

We might be able to use CMake to access the location of the Redistributable installer (https://gitlab.kitware.com/cmake/cmake/issues/17725).

chrisdembia:
Fixes #744, #807 

### Brief summary of changes

- Use CMake to find the location of the Visual C++ redistributable installer.
- Use Ant to place the redistributable installer within OpenSim's installation.
- Use NSIS to run the redistributable installer during the OpenSim installation.

### Testing I've completed

- Built the GUI and the installer locally and ensured that the Visual C++ redistributable installer runs when running the OpenSim installer.

### CHANGELOG.md (choose one)

- no need to update because...this restores previous behavior.

### TODO

- ~The NSIS installer seems to install everything twice. All files are unzipped twice, and the Visual C++ redistributable is invoked twice. I tried googling to see if this is a common issue with NSIS but I didn't find anything. @aymanhab, any ideas?~ EDIT: This is now fixed.

### DIFF
--- a/Gui/opensim/CMakeLists.txt
+++ b/Gui/opensim/CMakeLists.txt
@@ -23,8 +23,19 @@ add_custom_target(CopyOpenSimCore ALL
                   WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
                   COMMENT "Copy opensim-core Java Bindings to GUI.")
 if (WIN32)
+    # InstallRequiredSystemLibraries defines MSVC_REDIST_DIR,
+    # which is not really intended for use outside of 
+    # InstallRequiredSystemLibraries, but it's the best mechanism
+    # we have for now.
+    # https://docs.microsoft.com/en-us/cpp/ide/deployment-in-visual-cpp
+    # https://gitlab.kitware.com/cmake/cmake/issues/17725
+    include(InstallRequiredSystemLibraries)
+    # TODO: We only support 64-bit now. If building for 32-bit, must
+    # use vcredist_x86.exe (could use CMAKE_MSVC_ARCH).
+    set(MSVC_REDIST_EXE "${MSVC_REDIST_DIR}/vcredist_x64.exe")
+
     add_custom_target(PrepareInstaller ALL
-                  COMMAND ${Ant_EXECUTABLE} copy-installer-files -Dapi.dir="${OpenSim_ROOT_DIR}"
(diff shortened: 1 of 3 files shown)


chrisdembia:
@aymanhab the files are only included in the installer once. I think the installer is just running the "script" twice.

chrisdembia:
Maybe the `/norestart` flag is relevant? 

http://voidknot.blogspot.com/2013/12/shipping-visual-studio-redistributable.html

chrisdembia:
Nevermind...the installer installs twice even if I comment out the new line related to the redistributable installer.

Maybe the issue is with the NSIS script itself.

aymanhab:
This maybe relevant 

http://nsis.sourceforge.net/Reference/SetCompressor

chrisdembia:
I just tested an older installer (OpenSim-a7059a50-2018-05-08-win64.exe) and it also runs the installation twice. So the installing-twice issue is unrelated to this PR. I've created a new issue (#807).

chrisdembia:
I figured out what the issue is; I'll update this PR soon.

aymanhab:
LGTM, thanks for the find @chrisdembia At some point we may want to separate GUI from API, Doxygen documentation as separate components but for now no need for components indeed. Feel free to merge. travis build is due to out of date SWIG generated files. Not sure why travis only though.

chrisdembia:
Thanks @aymanhab . I'll merge this.

> At some point we may want to separate GUI from API, Doxygen documentation as separate components but for now no need for components indeed. 

For sure; I am not opposed to having components (that's why I didn't completely delete the line). It's just that at this point we only have 1.

>  Not sure why travis only though.

Because the issue exists only for non-PR builds and we don't have a non-PR build for AppVeyor (AppVeyor allows us to only have the PR build).

Issue #628, opened 2018-03-26
Latest intaller (03-26) uses 3.3 style application icon instead of 4.0. 


aseth1:
### Steps to reproduce
Install OpenSim build 0326_01 from [here]( https://www.dropbox.com/home/DevBuilds/build0326_01)

### Expected result
![image](https://user-images.githubusercontent.com/6494122/37929471-eb9736a0-30f4-11e8-9401-b5b3830caf22.png) both on the opensim64.exe in explorer and on application dock.

### Actual result
![image](https://user-images.githubusercontent.com/6494122/37929414-c8a057f8-30f4-11e8-9a56-eeb8f8bea61e.png)

### Environment and GUI version
Listed above.

aymanhab:
We don't have an "ico" file with the "4.0" style, we have to have these in sizes 48x48, 32x32, 16x16

jenhicks:
![opensimlogowindows_48](https://user-images.githubusercontent.com/4510959/37931238-dba8eb4e-30f9-11e8-9ca7-d78b00028420.png)

We had tried adding an outline to the logo, but perhaps making the jumper dude grayscale would look better

jenhicks:
@aymanhab I just emailed you some new icons. Let me know if there are issues.

aymanhab:
Ok, thanks @jenhicks though not showing up yet, maybe Dropbox easier?

jenhicks:
I emailed you a dropbox link

aymanhab:
Fixed in latest posted installer, thanks @jenhicks for the icons

aseth1:
I don't see the fix. There is no application icon associated (just default provided by netbeans) with OpenSim. This on Windows installer OpenSim e07981e0-2018-05-15:

![image](https://user-images.githubusercontent.com/6494122/40323882-a8469b7a-5ceb-11e8-89d4-b9fbbb65f713.png)

aymanhab:
@aseth1 the change of the icon is done using a tool that's not integrated into the automated build process. @chrisdembia was working on the integration on appveyor last week. Not sure where it stands but worst case we can do this manually for actual release builds.

chrisdembia:
I started working on this at the end of last week. I hope to find some time to continue working on it this week but no guarantees.

aseth1:
I think this can be easily verified on a manual build, so @chrisdembia don't worry about it while you are away.

chrisdembia:
[...]

PR #811, opened 2018-05-31
Set icon for OpenSim64.exe. Fix redist installer.

### LINKED ISSUE #809: Installer hangs for ~1 minute with progress bar at 100%
### Steps to reproduce
Install OpenSim-e1c6a4db-2018-05-30-win64.exe.

### Expected result
Installer provides feedback to user about progress of install.

### Actual result
New behavior not seen in installer from earlier today: after the progress bar reached 100%, a Windows dialog something like [this](https://msdn.microsoft.com/dynimg/IC726238.png) appeared. After clicking *OK*, the Windows dialog disappeared and nothing seemed to be happening for ~1 minute—long enough that I think many users would have [killed the process](https://i.redditmedia.com/_PSxMEdL3NulmUYRFndAKSk2SlLHl_tMYJ98GpMPtQA.png?w=745&s=c969218ac3413f24180451710f3ce8af).

OpenSim-e1c6a4db-2018-05-30-win64.exe

chrisdembia:
Partially addresses #628.

Fixes #809.

### Brief summary of changes

- We use ReplaceVistaIcon.exe (similar to Resource Hacker) to change the icon for OpenSim64.exe
- The NSIS installer now creates start menu shortcuts.
- The installation of the Visual C++ redistributable now occurs in a way wherein users are not likely to kill the installer (a separate installer pops up).
- I added the ability to set the icon for the installer executable itself, but I commented this out because we do not have an icon that is appropriate for the installer executable (I don't want to use the OpenSim logo on its own, because then the installer could be confused with the application itself. Ideally we would have the modern OpenSim in a cardboard box; similar to OpenSimInstallerIcon.bmp).

### Testing I've completed

- Built the GUI and installer and ran the installer.
- Ensured that the OpenSim desktop and start menu shortcuts work as intended, and that they have the correct icon (the OpenSim logo).
- Ensured that the process of running the Visual C++ redistributable is less confusing.

### CHANGELOG.md (choose one)

- no need to update because...minor changes


chrisdembia:
Yes this is ready to be merged.

Issue #766, opened 2018-05-19
User is not prompted before overwriting existing model file


tkuchida:
### Steps to reproduce
- open arm26.osim
- select *File > Save Model...* (accidentally---instead of the *Close All* item above it)

### Expected result
User is prompted before the existing arm26.osim file is overwritten.

~~Also note that the ellipsis following a menu item is typically used to indicate that the user will be prompted for additional information (e.g., destination directory, filename) before performing an operation. There should not be an ellipsis following "Save Model" (even if the user is asked to confirm).~~ [Edit: opened Issue #769]

### Actual result
GUI silently overwrites existing file. The old model file is lost!

### Environment and GUI version
d8bfa962-2018-05-09

aymanhab:
Tom, it's very typical if a model comes from a file and you're saving to overwrite old file. Same as all other software e.g word. If you do save as then you should be prompted. If model didn't have a file as in scaling output you should be prompted. Behavior is same in 3.3 but open to ideas for improvement.

tkuchida:
Please let me know how the user experience has been designed for the following use case:

User has many model files in version 20303, one for each subject in several experimental conditions. Diffing these files reveals the unique properties for each subject, and user has developed other infrastructure that works with model files in version 20303. User opens one of the model files and deletes a model marker. The "undo" button remains disabled so the user proceeds to close and re-open, but accidentally saves. The model file is now in version 30516; there are thousands of differences. User is unable to undo the conversion.

tkuchida:
> Same as all other software e.g word.

I don't think OpenSim is providing users with the typical feedback. Here's a use case:
- open arm26.osim from Resources directory (version 20303)
- *File > Close Model*

The model closes without prompting the user to save changes. The user's mental model tells them that a "save" operation would not have had any effect because, if there were any changes, the user would have been prompted to save changes before closing. Now:

- open arm26.osim from Resources directory (version 20303)
- *File > Save Model...*
- *File > Close Model*

The model file is different (now version 30516). Telling the user "it's your own fault for not saving a backup copy" or "it's your own fault for clicking *Save Model*" or "you need to update your existing infrastructure if you're using 4.0" makes for a very negative user experience.

I would expect that many users will open their existing model files to try out the new version of OpenSim. Clicking *Save Model* performs an irreversible change to the file which could negatively affect their research. That could be their first interaction with OpenSim 4.0.

[...]

PR #879, opened 2018-07-03
Warn users if saving over a Pre 4.0 model file

aymanhab:
Fixes issue #766 
### Brief summary of changes
If model is detected to have come from a Pre 4.0 format, a dialog is shown to warn and ask for confirmation to overwrite or to cancel and use SaveAs instead

### Testing I've completed
tested saving over old model files
### CHANGELOG.md (choose one)

-Need to update to indicate the reason.


aymanhab:
Appreciate feedback on the dialog which looks like this now:
capturewarnversionchange
 The dialog shows on save, if user cancels nothing happens, while ok overwrites the model file.

jenhicks:
I would put a question mark after "Overwrite old model file?"

Issue #643, opened 2018-04-02
Change default marker size for model and experimental markers to 10mm


aseth1:
### Steps to reproduce
Preview or associate marker data in the GUI.
Click on the *Markers* under *Motions* in the Navigator.
*marker color* and *marker size* are listed as properties.
Changing the color has an effect on the visualized experimental markers but size has no effect.

### Expected result
*marker size* number reflects the size of the marker

### Actual result
No effect.
<img width="1437" alt="screen shot 2018-04-01 at 7 12 53 pm" src="https://user-images.githubusercontent.com/6494122/38180458-9950c96e-35e1-11e8-8688-e5de26b45a3b.png">

*marker radius* would also be more clear and its units should be *m*.

### Environment and GUI version
Mac OpenSim 651d29f-2018-03-30

aymanhab:
This is practically a duplicate of #474 since analytical shapes do not respond to scale changes. Will be addressed together.

jenhicks:
Thanks @aseth1 @aymanhab ... keeping issue open so that we can verify this use case is also fixed.

aseth1:
Unlike #474, value changes are accepted and *size* does not return to 1.0. Also not obvious that the issues are related.

tkuchida:
Current behavior is unexpected and internally inconsistent.

**Steps to reproduce:**
- File > Preview Experimental Data...
- Select test.trc (contained in this ZIP file: [test.zip](https://github.com/opensim-org/opensim-gui/files/1931234/test.zip))
- Can't find markers in Visualizer (Issue #653)
- File > Close All
- File > Preview Experimental Data...
- Select test.trc
- In Navigator, select *Markers* and change `marker size scale`

**Expected behavior:**
Marker size is absolute. Changing from 1.0 to 2.0 makes markers twice as large. Changing again from 2.0 to 2.0 (i.e., editing the field but not changing the numbers) should have no effect. Changing again from 2.0 to 1.999 should make no noticeable difference.

**Actual behavior:**
Marker size is relative *and* you are not permitted to make the same relative change in the marker size twice. Changing from 1.0 to 2.0 makes markers twice as large. Changing again from 2.0 to 2.0 (i.e., editing the field but not changing the numbers) does nothing---which would be expected if the size were absolute. However, now changing from 2.0 to 1.999 makes the markers (just less than) twice as large. Changing 1.999 to 2.0 doubles the size of the markers again.

[...]

PR #887, opened 2018-07-12
Fix experimental marker preference

aymanhab:
Fixes issue #643

### Brief summary of changes
Introduce new preference "Experimental Marker Size (mm)" that specifies size in millimeters. 

### Testing I've completed
Previewed Marker data, changed preference, reloaded, restarted, Preference and behavior is as expected.

### CHANGELOG.md (choose one)

- Need to document new preference

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/actions/EditPreferencesJPanel.java
+++ b/Gui/opensim/view/src/org/opensim/view/actions/EditPreferencesJPanel.java
@@ -74,7 +74,8 @@ public class EditPreferencesJPanel extends javax.swing.JPanel {
        "Muscle Dsiplay Radius",
        "BuildDate",
        "One Material Meshes",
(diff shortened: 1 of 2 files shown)


aseth1:
If the value is now the absolute dimension of the marker the name should reflect the dimension being set. Is size really the "Experimental Marker Radius"?

This should not have been merged without review.

aymanhab:
@aseth1 The value now is correctly in mm,  issue #643 was about the behavior rather than the name of the preference and this has been fixed. I agree "radius" is more accurate if it remains a sphere but in the past and possibly in the future we can allow other shapes (e.g. cubes). Please feel free to open an issue for the name as you see fit.

Ideally we can make GUI builds off branches to test/iterate/get feedback but as of now that's not possible. Thank you.

Issue #894, opened 2018-07-17
Saving transformed experimental data erroneously rotates data by 180 degrees


aymanhab:
### Steps to reproduce
Preview Experimental data , expand node in navigator and choose Transform from context menu

### Expected result
Dialog to allow for transforming the data and saving it.

### Actual result
Exception thrown

(include screen grabs where appropriate)

### Environment and GUI version
4.0 stream

jenhicks:
Please also test whether data is transformed as expected when saved and reloaded

aseth1:
The data appears to be transformed correctly. But when saved and reopened the data is 180degs off.
![image](https://user-images.githubusercontent.com/6494122/43173036-73cb5846-8f67-11e8-8ae8-69c6f61c922d.png)

jimmyDunne:
Dito on Mac.

nickbianco:
I observed the proper transformation behavior via #911.

jimmyDunne:
Transform is correctly saved on Mac 🍾

PR #896, opened 2018-07-18
Transform experimental data

aymanhab:
Fixes issue #894

### Brief summary of changes
moved code that transforms experimental data from using vtk objects/transforms to using SimTK::Transform iternally and communicating changes to visualizer using messages.

### Testing I've completed
Was able to load marker data, forces, rotates them, write down and reload transformed/rotated data.

### CHANGELOG.md (choose one)

- no need to update because this is a bugfix that restores functionality in earlier versions of OpenSim.

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/experimentaldata/AnnotatedMotion.java
+++ b/Gui/opensim/view/src/org/opensim/view/experimentaldata/AnnotatedMotion.java
@@ -44,6 +44,7 @@ import org.opensim.modeling.ArrayStr;
 import org.opensim.modeling.Model;
 import org.opensim.modeling.StateVector;
 import org.opensim.modeling.Storage;
+import org.opensim.modeling.Transform;
 import org.opensim.view.motions.MotionControlJPanel;
 import org.opensim.view.motions.MotionDisplayer;
 import org.opensim.view.motions.MotionsDB;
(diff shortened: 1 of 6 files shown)


Gui/opensim/view/src/org/opensim/view/experimentaldata/AnnotatedMotion.java
```diff
@@ -290,23 +292,28 @@ public void setBoundingBoxComputed(boolean boundingBoxComputed) {
...
  
-    public Storage applyTransform(vtkTransform vtktransform) {
+    public Storage applyRotations(Vec3 rotationAnglesInRadians) {
         // Should know what data to apply xform to and how
         // for example mrkers are xformed as is but force vectors would not apply translations etc.
         // utilize the "clasified" table
         Storage motionCopy = new Storage(this);
+        Transform simtkTransform = new Transform();
+        simtkTransform.R().setRotationToBodyFixedXYZ(rotationAnglesInRadians);
+        Mat33 mat33 = simtkTransform.R().asMat33();
```
aseth1: mat33 is unused and should be removed.

aymanhab: Agreed, thanks @aseth1 👍

Issue #900, opened 2018-07-23
Update help links in the tools and other menus to point to 40 version of the online documentation.


jenhicks:
The help buttons in the tools (Scale, etc.) are still pointing to 3.3 versions.
The help button for plotter, excitation editor, preferences, still points to 3.3

If there is a local redirect, these should also be removed, as the doc is all online.

jenhicks:
All looks good on my Mac. Thanks for making this fix @jimmyDunne !

PR #902, opened 2018-07-24
Tool Help URLs to 4.0 Docs

jimmyDunne:
Fixes issue #900

### Brief summary of changes
Changed URL to point to 4.0 Docs

### Testing I've completed
URLS go to the correct locations

### DIFF
--- a/Gui/opensim/helpUtils/src/org/opensim/helputils/helpmenu/OpenOnlineDocsAction.java
+++ b/Gui/opensim/helpUtils/src/org/opensim/helputils/helpmenu/OpenOnlineDocsAction.java
@@ -32,7 +32,7 @@ import org.opensim.utils.BrowserLauncher;
 public final class OpenOnlineDocsAction extends CallableSystemAction {
 
     public void performAction() {
-       BrowserLauncher.openURL("https://simtk-confluence.stanford.edu:8443/display/OpenSim/OpenSim+Documentation");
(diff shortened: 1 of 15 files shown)


jenhicks:
Thanks @jimmyDunne 

The links look right. Two comments:

- We should probably do a search of the GUI code to make sure there aren't any other OpenSim30 links floating around. Is that possible @aymanhab 
- The code is still doing a check to see if the user is online and trying to send to a local copy if the user is not online. For these pages, we are not distributing a local copy, so we may need to remove or update the code accordingly.

aymanhab:
captureopensim30
File names and line numbers shown. Good idea @jenhicks 
@jimmyDunne You should modify the code as isConnected() is set to true, and an empty page-not-found-notOnline page will show up.

aymanhab:
Exactly.

aymanhab:
@jimmyDunne The changes look good but the editor you used messed around with newlines/spaces causing artificially big diffs in some files. if you transplant the changes using a text editor that doesn't try to reformat that would be great, then we can merge this PR. Thank you.

The *JNI files are autogenerated so we need to make a change in opensim-core for this change but this is done in a separate repo/PR.

chrisdembia:
@aymanhab you can use this link to view the diff without changes in whitespace: https://github.com/opensim-org/opensim-gui/pull/902/files?w=1

aymanhab:
Thanks for the pointer @chrisdembia, I changed the diff options temporarily and that's how I figured the changes were good but still the reformatting would be inconsistent when editing these files later.

Issue #512, opened 2018-02-11
Undo button remains disabled following undo-able action


tkuchida:
**Steps to reproduce**
Open *arm26.osim* and select Bodies -> Display -> Hide. Bodies disappear but undo button remains disabled. Editing a property (e.g., mass of a body) enables the undo button, whereupon the property edit as well as the previous hide operation can be undone.

**Environment**
Windows, with *OpenSim-078b8daf-2018-01-26.zip* artifact from AppVeyor.

carmichaelong:
Notes from #909 have been accounted for and placed into separate issues

PR #909, opened 2018-07-26
Enable undo

aymanhab:
Fixes issue #512
### Brief summary of changes
Enablement of the undo button is handled by the NetBeans platform based on what windows (TopComponents) are "active" and if hey provide an undo support class. This PR unifies all editing to make one window Active and provide a singleton Undo support class across application windows.

### Testing I've completed
Edits in Properties window, context menus of the Navigator and graphically dragging PathPoints are all undoable and the undo button is enabled.

### CHANGELOG.md (choose one)

- no need to update because it is a bugfix

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/ExplorerTopComponent.java
+++ b/Gui/opensim/view/src/org/opensim/view/ExplorerTopComponent.java
@@ -89,7 +89,7 @@ final public class ExplorerTopComponent extends TopComponent
    private final BeanTreeView modelTree = new BeanTreeView();
    
    private static TopComponent visualizerTC = null;
-   private UndoRedo.Manager undoRedoManager = new UndoRedoManager();
+   private UndoRedo.Manager undoRedoManager = null;
 
    private ExplorerTopComponent() {
       initComponents();
@@ -471,6 +471,7 @@ final public class ExplorerTopComponent extends TopComponent
         }*/
         
         // Undo Support
+   @Override
         public UndoRedo getUndoRedo(){
             return getUndoRedoManager();
 }
@@ -535,22 +536,16 @@ final public class ExplorerTopComponent extends TopComponent
         
         public static void addUndoableEdit(AbstractUndoableEdit aUndoableEdit)
         {
-            getDefault().getUndoRedoManager().addEdit(aUndoableEdit);
-            TopComponent tc = getVisualizerTopComponent();
-            if (tc==null){ // No gfx window
-                tc = getDefault();
-            } 
-            if (tc==null) return;   // No tc to piggyback on
-            final TopComponent tcf = tc;
+            instance.getUndoRedoManager().addEdit(aUndoableEdit);
             if (java.awt.EventQueue.isDispatchThread()) {
-                tcf.requestActive();
+                instance.requestActive();
             }
             else {
                 SwingUtilities.invokeLater(new Runnable(){
 
                 @Override
                 public void run() {
-                    tcf.requestActive();
+                    instance.requestActive();
                 }
             });
             }


aymanhab:
@chrisdembia Not sure if review would make sense since these are bits and pieces of a larger context. @jimmyDunne , @tkuchida or @aseth1 if you can test the artifact of this PR on Windows and give me feedback that would be great. Thank you.

carmichaelong:
Testing with arm26 on OpenSim a6d9c27e-2018-07-27:

Below are the testing steps I took, trying a number of use cases. I have highlighted some issues I ran into and marked them with **(Needs discussion)** or **(Needs fix)** or some combination of both when I wasn't sure which was more appropriate.

**Hiding bodies**
a) Bodies -> Display -> Hide.
b) Bodies, r_humerus -> Display -> Hide.
- **(Needs discussion)** Similar results: Undo works, but needs multiple clicks to work as might be expected. For instance, there are 4 objects within r_humerus, and 4 undos are needed to go back to the original state.

**Changing color**
a) Bodies, r_humerus, r_humerus_geom_1: changing color
- Undo returns back to the original color.
b) r_humerus_geom_1: change color 4 times
- Undo steps through each color change.

**Show axes or show mass center**
a) r_humerus: Show axes
b) r_humerus: Show mass center
c) Joints, r_shoulder, base_offset: Show axes
- **(Needs discussion)**: These act like toggles as previously, but these changes do not trigger an undo.

d) r_humerus -> show axes, then r_humerus_geom_1 -> Hide
- Undo will do the "unhide" step and does not affect the axes shown.

**Muscle properties**
All tests done by changing the value in the properties window, hitting the undo button, and seeing the property change back.
a) Tested one at a time changes (floats: max_isometric_force, optimal_fiber_length, tendon_slack_length, max_contraction_velocity, max_pennation_angle
b) Tested one at a time changes (bool): appliesForce
- Correctly reset to the original value in the model

**Marker location**
1. Change r_acromion location from (-0.01256 0.04 0.17) to (-0.01256 0.1 0.17). DO NOT CLICK ANYWHERE ELSE.
Result: Property updates correctly and moves in visualizer.
2. Hit Undo.
- **(Needs fix)** Result: Visualizer updates, but properties window does not update. This is distinctly different than what is observed with muscle properties.
3. Click in the Navigator.
- Result: The value in the property updates.

**Undo/Redo button behavior when working with any property**
1. Bodies -> base. Change the mass_center to (0 1 0).
Result: mass_center updates, and the Undo button lights up.
2. Click anywhere in the properties window:
a) On the top bar (i.e. anywhere here):
property-top-bar
b) On a property
- **(Needs discussion/fix)** Result: Undo/Redo buttons grey out and cannot be used.
3. Click anywhere out of the properties window
Result: Undo/Redo buttons come back and can be used

**Moving a path point graphically**:
1. TRIlat -> Show Only
2. Ctrl-click a path point and move it.
3. Hit Undo
- **(Needs fix)** Result: Muscle path is quite contorted.

**Joint offset translation/rotation**
1. Joints -> r_shoduler -> base_offset
2. Change Translation or Rotation
- **(Needs fix)**: Undo does not come up.

aymanhab:
Thanks for testing @carmichaelong Let's be clear on the scope of the fix: Collapsing multiple undos, or commands that have not been undoable in the past (COM, Axes that do not correspond to properties) do indeed need discussion if no issues are open already but are unrelated to the enabling of undo button. 

As I explained in some earlier meeting, the undo behavior has two components:
1. What operations to perform when undo/redo is actually chosen by user. 
2.  Enable the undo/redo button 
This PR deals only with 2. Individual issues would need to be opened for 1 as needed. 

With that in mind, I believe the only sequence you reported that falls under 2. is editing multiple Properties while focus stays in the Properties window. I appreciate if you'd confirm and I'll investigate. Thank you :+1:

carmichaelong:
> If you hit enter or tab (used across the application to indicate finish edit) after a property edit then you get the expected undo enabled.

While testing yesterday, this was true when I was testing all muscle properties. For some reason when testing today, this behavior is not true for any property at all (including muscle properties). I have to click back on the Navigator window to see the Undo option, and does work as expected in this PR at that point.

carmichaelong:
Separately, if I understand the scope of this PR, then it LGTM for merging, and some subset of the testing cases here should be moved to current or new issues. If we merge, then the test case should be something like:

1. Open up any model.
2a. Change any property (float or bool).
2b. Move a muscle path point graphically.
2c. Hide a body.
3. Click on the Navigator window.
4. Ensure that the Undo button shows up. 
5a. Undoing a property panel change should be done correctly, 
5b. Undoing a muscle path change will change the path, but it won't update back to the original path
5c. Undoing a "Hide" will only undo "one-by-one" but should work.

@aymanhab Let me know if this makes sense? I'm happy to discuss further.

aymanhab:
Perfect, thank you @carmichaelong much appreciated.

chrisdembia:
I repeated the test that @carmichaelong mentioned on Mac, OpenSim 95345a0-2018-08-02. So, this issue is *Verified*.

However, multiple property edits cannot be redone. I've opened a separate issue for that (#938).

Issue #869, opened 2018-06-27
Exception when using utils.FileUtils.getInstance().browseForFolder()


jimmyDunne:
Jen: Scripts in the GUI commonly ask users to browse for a folder. Need an example on Mac +Windows of how to do this.

Use code in GUI script:
`resultsFolderCMC = utils.FileUtils.getInstance().browseForFolder("Select the folder with CMC Results")`

Returns an exception;

```
>>> Traceback (most recent call last):
  File "/Users/jimmy/repository/opensim-models/Tutorials/Design_to_Reduce_Metabolic_Cost/Scripts/CMC_EvaluateResults.py", line 8, in <module>
    resultsFolderCMC = utils.FileUtils.getInstance().browseForFolder("Select the folder with CMC Results");
TypeError: browseForFolder(): expected 0 or 2-3 args; got 1
```
I suppose the method has changed. how is the new version written?

tkuchida:
Perhaps `utils.FileUtils.getInstance().browseForFolder(0, "blah blah")`?
https://msdn.microsoft.com/en-us/library/windows/desktop/bb774065(v=vs.85).aspx

jimmyDunne:
Thanks, Tom. Still something funky going on. 

`resultsFolderCMC = utils.FileUtils.getInstance().browseForFolder("Select the folder with CMC Results",0);` opens a dialogue, but when a folder is selected an exception is returned;

``` File "<input>", line 1, in <module>
	at org.opensim.utils.FileUtils.browseForFolder(FileUtils.java:286)
	at org.opensim.utils.FileUtils.browseForFolder(FileUtils.java:253)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)

java.lang.NullPointerException: java.lang.NullPointerException
```
I am doing this on a Mac. This seems to be an issue with the custom functions [here](https://github.com/opensim-org/opensim-gui/blob/6a65942e04327542e053fb6450d64b28fcbf1bf9/Gui/opensim/utils/src/org/opensim/utils/FileUtils.java#L246-#L290). @aymanhab, please let me know what the call should be or if it requires a fix on the GUI side. Thanks.

[...]

PR #920, opened 2018-07-30
Cache resulting folder name before changing dialog settings

aymanhab:
Fixes issue #869

### Brief summary of changes
When user makes selection of Folder on OSX, we use that as the new WorkingDirectory for future browsing. This minor refactoring avoids using selectedFile after changing the file browser settings (which wipes out the selected file on OSX but not windows)

### Testing I've completed
Launched browse for folder dialog from scripting shell on osx and selected folders without issue or null ponters
### CHANGELOG.md (choose one)


aymanhab:
Build failure seems bogus. Any explanation @chrisdembia

Issue #816, opened 2018-05-31
Graphically moving path point displays a convoluted path


aseth1:
### Steps to reproduce
Load arm26.osim. Select TRILat middle via point and drag.

### Expected result
Path drawing updates contiguous with existing path points.

### Actual result
Path point moves and is redrawn with respect to two phantom path points which remain stationary with respect to the other via points.
![image](https://user-images.githubusercontent.com/6494122/40813495-0630d974-64f0-11e8-98ca-1be33cbe77f1.png)

### Environment and GUI version
OpenSim-43c1be65-2018-05-30

jenhicks:
I see the same behavior, though if you wait a seconds, the muscle path is redrawn as you would expect.

aseth1:
Still present in Mac OpenSim 1e476f2-2018-06-13.
Discussed with @aymanhab and he felt that because this transitory (I am still dragging) that the kinks are not important since they disappear on release. I think it is goofy.

tkuchida:
> I think it is goofy.

![](https://user-images.githubusercontent.com/4203505/41374990-90fce7fa-6f09-11e8-8af1-c8c484431723.png)

Perhaps only the point (sphere) should be dragged and the path should update only once the mouse button has been released? That would also allow the user to see where they're moving the point in relation to the existing muscle path.

aymanhab:
@tkuchida Since pathpoints can't exist in multiple places at the same time, and they need to exist in the old location for the old path to be drawn, I don't think this solution is sustainable without major rework.  
We can try to update the visualization based on a timer (e.g. every .1 sec while dragging) but we need to prioritize relative to other issues.

aymanhab:
While the transient behavior is likely to stay, there's a bug where graphically moving pathpoitns doesn't follow the full protocol of updating path and recreating system as done for other editing in Dialog. Fix in progress.

aseth1:
On Windows OpenSim 11d7a130-2018-08-03 I still don't think the behavior is correct.
Starting off like this:
![image](https://user-images.githubusercontent.com/6494122/43670612-7f231534-9742-11e8-8ba3-5a601ed055dd.png)

Selecting points to edit seems to add a new path point and doesn't move the existing point! And unlike before it is not transitory - it persists. Undo does not work here either and undoes the hiding of muscles that were hidden rather than undo the current pathpoint edit.

[...]

PR #923, opened 2018-07-31
Use OpenSimContext setLocation method

aymanhab:
Properly edit GeometryPath point location and recreate the system, also refresh path display on edit/undo.

Fixes issue #816

### Brief summary of changes
In the past, graphical editing of pathpoint called set_location directly on the pathpoint which is incomplete since Path cache entry and system underneath weren't updated. This PR (along with opensim-core https://github.com/opensim-org/opensim-core/pull/2234) fixes this by making the right call sequence when doing graphical editing consistent with how it's done in path edit dialog.

### Testing I've completed
Can move pathpoints on GeometryPaths with wrapping and undo the changes

### CHANGELOG.md (choose one)

### DIFF
--- a/Gui/opensim/modeling/src/org/opensim/modeling/OpenSimContext.java
+++ b/Gui/opensim/modeling/src/org/opensim/modeling/OpenSimContext.java
@@ -218,7 +218,11 @@ public class OpenSimContext extends OpenSimObject {
   }
 
   public void setLocation(PathPoint mp, int i, double d) {
-    opensimActuatorsAnalysesToolsJNI.OpenSimContext_setLocation(swigCPtr, this, PathPoint.getCPtr(mp), mp, i, d);
+    opensimActuatorsAnalysesToolsJNI.OpenSimContext_setLocation__SWIG_0(swigCPtr, this, PathPoint.getCPtr(mp), mp, i, d);
+  }
+
+  public void setLocation(PathPoint mp, Vec3 newLocation) {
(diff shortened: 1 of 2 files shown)


aymanhab:
This is WIP pending merging the opensim-core side of the fix since it uses it, builds will fail till then.

chrisdembia:
@aymanhab thanks for your comment in https://github.com/opensim-org/opensim-core/pull/2234 but I still don't understand how this fixes #816. I suppose I will wait until I can test this locally.

chrisdembia:
> In the past, graphical editing of pathpoint called set_location directly on the pathpoint which is incomplete since Path cache entry and system underneath weren't updated. This PR (along with opensim-core opensim-org/opensim-core#2234) fixes this by making the right call sequence when doing graphical editing consistent with how it's done in path edit dialog.

Does this cause the visualization to be updated as soon as a user graphically moves a path point?

aymanhab:
@chrisdembia you should be able to test visual editing and undo for muscles including wrap objects. While path remains "goofy" during dragging, one done, path should restore to expected and same with undo. Please let me know if you can test this out. Thank you.

chrisdembia:
Verified on Mac: can properly undo edits to path points.

aug-02-2018 17-25-39


Does this also address #928 ?

aymanhab:
Thanks much @chrisdembia I believe it does address issue #928 as well

Issue #895, opened 2018-07-17
Movie/video maker does not respect playback speed


aseth1:
Jen: Users expect to be able to make beautiful movies to illustrate their simulations. Being able to change playback speed is a key part of this.

### Steps to reproduce
1. Load the drop landing model from [here](https://stanford.box.com/s/3ifyk4w9py6lbxltq75jxx62c4n72ae7) 
2. load the original states file: [incline_30.0_gain_10.0_delay_0.06_states .sto.txt](https://github.com/opensim-org/opensim-gui/files/2203714/incline_30.0_gain_10.0_delay_0.06_states.sto.txt)
3. The total landing motion is only 0.15s, so set the playback speed to 0.1 to animate in slow motion.
4. click movie camera icon to start recording, press play and then click camera to stop recording.

### Expected result
Resulting video is 1.5s in duration with some 45-90 time frames (30-60Hz).

### Actual result
The resulting video lasts only 0.02s and contains 3 time frames.

### Environment and GUI version
Windows OpenSim_b435c8fa-2018-07-06

aymanhab:
Some notes:
1. the motion associated with the model is actually 15 seconds rather than 0.15 seconds . Since we try to play motion real time by default, using very small time is fundamentally error prone (how soon did you push the buttons to start/stop etc. what's the startup time to initialize the capture etc.).
Using the 15 sec motion out of the box produces videos of length:
 - normal speed: 15 sec.
 - half speed: 26 sec.
 - tenth of speed: 2 min 2 sec.

2. I did tweak some parameters on the capture, and was able to get 15, 30, 2:28 so I can make a build with this change.

Not sure how you counted frames in webm to tell there're 3 frames but would be good to know so we're in sync.

-

aseth1:
> the motion associated with the model is actually 15 seconds rather than 0.15 seconds .

I manually lengthened the time from 0.15s to 15s so I could produce a reasonable video. I believe I submitted the original file that was not stretched out. I just downloaded and it is 0.15s, so I don't know how you got the 15s file. There is no way the drop landing is naturally 15s.

I counted 3 frame by playing by sliding the MP4 player slider and saw it tansition between three frames.

aymanhab:
@aseth1 The zip file available from the link you posted at the top of this issue has only one sto file that has "time_stretched" in the name and indeed it is 15 seconds. Where do I get the non time stretched version?

aseth1:
Yes, but please read step 2!

[...]

Issue #299, opened 2017-07-16
Saving high-resolution screenshots from the visualizer cuts off the image


chrisdembia:
Jen: Key feature of OpenSim 4.0 is being able to create great visuals. Controlling the resolution of the image is a key part of this.

Here's how we could temporarily increase the size of the renderer while taking a screenshot: https://stackoverflow.com/questions/27534206/three-js-screenshot-custom-size-using-an-offscreen-rendere

It would be great to save screenshots as 300 dpi or maybe even 600 dpi.

aymanhab:
I implemented the solution suggested and it creates much nicer pngs indeed but you have to specify dimensions (in pixels) 
![opensim_snapshot_1830x1186](https://user-images.githubusercontent.com/3972208/33912602-6d6b4712-df4b-11e7-9e4a-e618688c85ea.png)
![opensim_snapshot](https://user-images.githubusercontent.com/3972208/33912605-73cb075a-df4b-11e7-9b5e-cffb4c0508ff.png)

Still deciding on entry point to avoid clutter. Not exactly DPI spec. but DPI has many players involved and no cross-platform measure it seems.

chrisdembia:
Do you mean the end-user must specify the dimensions? Why is that?

aymanhab:
The mapping between resolution DPI (misleading)/PPI and inches doesn't seem to be available in a reasonable (non-hacky way). The two options I'd suggest would be either ask user for size in inches and PPI and compute resolution from that (to pass to the code) or just have the user do the multiplication and specify resolution. 
https://www.iprintfromhome.com/mso/understandingdpi.pdf
Either way the function I implemented based on your link above seem to print at arbitrary resolution. Please let me know what you think.

chrisdembia:
Ah thanks for explaining; I think I understand. Having the user specify the dimensions in inches and then a separate resolution (in pixels/inch) sounds reasonable, though using inches might be confusing for those outside the US. Having the user specify the dimensions in pixels also sounds fine to me.

My main concern is about preserving the aspect ratio. We should make it hard for the user to change the aspect ratio.

aymanhab:
As of now I added a shortcut that when you press 'd' for high-Dpi in 3D window you get upsampled screen capture by a factor of 4. Would be good to know if this works cross platform and on a regular resolution monitors and if the result meets common journal standards  which usually require min DPI but no max/file-size.

aseth1:
Consider adding a user preference to set the the resolution factor.

[...]

PR #936, opened 2018-08-02
Bring fix to movie capture and default snapshot quality from threejs

aymanhab:
Fixes issue #895 and maybe #299

### Brief summary of changes
Two fixes in threejs repo are brought over:
-  Use different encoder and higher frame rate for movie capture
- Make snap_shot quality slider correspond correctly to value used internally so at least low res grab is possible

### Testing I've completed
More frames and smoother movies for model/motions in issue #895, resolution of png files captured map to quality slider as expected.

### CHANGELOG.md 


aymanhab:
@aseth1 you can try the new movie encoding in this PR and let me know what you see. Also the build has the fix to the snapshot quality slider, so at least you should be able to get un-cropped snapshots at low slider setting

aymanhab:
There's the issue 895 model and data files which was particularly short (0.15 sec.) I deally for a few models/motions of reasonable length we can check the length of playback time to see if it maps reasonably close to realtime (e.g. 2 sec, played at a speed of .1 produces roughly 20 sec of smooth motion etc.)

aseth1:
The artifact for this PR OpenSim_9551139f-2018-08-02 does not have the *snapshot quality slider* and the snapshot feature seems to be disabled.

image

aseth1:
As for video/movie recording it does seem better than I can capture 0.25s of motion at 0.1s playback and it occupies about 4s of movie time. However it includes about 1s of time for me to click from the 🎥 to the ⏯ buttons and then back to the 🎥 to stop recording. I don't think the video should be recording when there is no motion. Also it is impossible to discern if the video camera is in record mode. In 3.3 we had an orange highlight. Could we have some change in the icon or highlighting to indicate that recording is on?

aymanhab:
Thanks for the testing @aseth1 Can you check if the snapshot slider opens if you open the artifact for the second time (if you still have it around)? I'll undo that change to snapshot anyway so as to get the other fix to master and in the hands of more users. The behavior of when to start/stop recording should be the same as 3.3 though not optimal. When selected the background turns gray but that doesn't help with gray background, please open an issue and will try solutions.

aymanhab:
I undid the snapshot change just in case, if anybody can try the Artifact and confirm snapshot button/slider is available then will merge and move to verified on Windows. Thanks

aseth1:
Using the same artifact on Windows OpenSim_9551139f-2018-08-02 with the slider now present I can generate snapshots and quality 1 (556KB) and 2 (1602KB) and at 3 (3,049KB) and 4 (5952KB) the zoom changes and the view model get cutoff.
Original Screen Grab (365KB):
image

At quality 2:
opensim_snapshot_2

At quality 4:
opensim_snapshot_4

Hard to see without downloading the image and zooming, in but some of the muscles in front of the ankle (e.g. *per_tert_r*) do not appear clearly but only have strips drawn (this happens at all qualities, but not in the direct screen capture).

Issue #796, opened 2018-05-28
Default pathpoint scale up should be 1.05


jimmyDunne:
Previous OpenSim versions had the muscle pathpoints coloured in a slightly different shades of red and the sphere was distinguishable from the muscle. Path points are essentially indistinguishable from the muscle path. 

**OpenSim 4.0**
![muscle_paths_40](https://user-images.githubusercontent.com/5296507/40629896-f6830df4-6282-11e8-9636-b2d20e88e9e2.png)

**OpenSim 3.3**
![muscle_paths_33](https://user-images.githubusercontent.com/5296507/40629897-f88ca88a-6282-11e8-873d-4f10428aef06.png)

I don't necessarily have an issue with this, but it could now be easy to miss a path point if the deviation of the new path is very slight. It also doesn't make intuitive sense that this is a path point now- it looks as if the muscle path is just kinked. 

There feels like there is an asthetic vs function trade off here so it should be discussed. 

### Environment and GUI version
W10
OpenSim-543e2ecd-2018-05-27-win64.exe

tkuchida:
Could make the radius of the path point spheres slightly larger than the radius of the muscle cylinders, and display only when the muscle is selected.

aymanhab:
They are in fact kinks, except you have a handle to drag them. Larger radius will make ends look a bit worse though may improve z-fighting issues. How large considering size is user specified? 5%, 10%? Would be expensive to do on selection or to keep multiple geometries around for same pathpoint.

aymanhab:
@jimmyDunne @tkuchida in latest devBuild 0608, there's a separate command to toggle pathpoints to bigger size that should make it much easier to move/edit/grab. you either activate by doubleclick in navigator or explicit selection of command.

jenhicks:
On Mac, delete preferences

aseth1:
What is the default `PathPoint Scaleup` supposed to be?  Do we expect users to know what it means?

Setting it to a small (or large) value looks quite ridiculous.
![image](https://user-images.githubusercontent.com/6494122/43173702-2601f4ae-8f6b-11e8-8f04-3256a4e43805.png)

If the point is that it useful for selection, then they should always be of selectable size and should not be a user setting. For the time being. Setting the scaleup to1.05 appears to be reasonable (similar in appearance to 3.3).

Maybe the preference should just be a flag to make path points visible? Ideally the path points only appear when the muscle is selected and with different (highlight) color.

[...]

PR #937, opened 2018-08-02
Fix scaleup factor for  path points to 1.05 

aymanhab:
Fixes issue #796

### Brief summary of changes
Removed use of preference and fixed the number used to scaleup pathpoint to 1.05, stale preference will be handled with issue that handles new preferences in 4.0

### Testing I've completed
Loaded model, got reasonable sized pathpoints similar to 3.3, changed preferences had no effect on reloading

### CHANGELOG.md (choose one)

- no need to update because restores old behavior


chrisdembia:
I tested this on Mac and the path points look good:

image

As expected, the preference is gone (I ran `defaults delete org.opensim.utils` before starting the GUI):

image

PR #945, opened 2018-08-06
Sync experimental Marker scale property with visualization

aymanhab:
Fixes issue #643 

### Brief summary of changes
Experimental Marker size/color were being lost on switching current motion (or sync.) Now during initialization/switching we use size that accounts for both DefaultSize and scaleFactor

### Testing I've completed
Loaded multiple sets of experimental data (Markers) and was able to switch/sync motions and the size stays in sync with property

### CHANGELOG.md (choose one)


aymanhab:
Ready for testing/verification @nickbianco when you have a chance. Thank you 👍

nickbianco:
Verified on my machine that the `marker scale size` property stays in sync with Experimental Data object after syncing motion or re-selecting after un-selecting (Windows 7, OpenSim d53bd5be-2018-08-06).

aymanhab:
Updated comment to trigger a rebuild of gui with latest opensim-core. Builds of this PR should have the complete fix. Note that on Windows the Preferences are persistent while on Mac they're recreated every time.

aymanhab:
@nickbianco @jimmyDunne @carmichaelong if you can get the artifact from the PR on travis and test that would great. Thank you.

Gui/opensim/view/src/org/opensim/view/experimentaldata/MotionObjectPointForce.java
```diff
@@ -180,7 +180,7 @@ public void generateDecorations(boolean fixed, ModelDisplayHints hints, State st
             Transform xform = new Transform();
             
             xform.setP(new Vec3(point[0], point[1], point[2]));
-            appendToThis.push_back(new DecorativeArrow(new Vec3(0, 0, 0)).setBodyId(0).setColor(new Vec3(0., 1., .0)).setOpacity(0.5).setIndexOnBody(getStartIndexInFileNotIncludingTime()).setTransform(xform));            
+            appendToThis.push_back(new DecorativeArrow(new Vec3(0, 0, 0)).setBodyId(0).setColor(color).setOpacity(0.5).setIndexOnBody(getStartIndexInFileNotIncludingTime()).setTransform(xform));
```
aseth1: This line of code is indecipherable. Please please describe what this is supposed to be doing? `setBodyId(0)...setIndexOnBody(getStartIndexInFileNotIncludingTime())` is very confusing. What is `getStartIndexInFileNotIncludingTime()` and why is it relevant or necessary in this context?

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
```diff
@@ -376,7 +375,7 @@ public Vec3 getDefaultForceColorVec3() {
     
     /** Creates a new instance of MotionDisplayer */
     public MotionDisplayer(Storage motionData, Model model) {
-        String saved = "5";
+        String saved = "10";
```
aseth1: Please rename the variable to what is supposed to represent, e.g. `defaultMarkerRadius`

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
```diff
@@ -988,7 +987,7 @@ private void createDefaultMotionObjects() {
             UUID uuidForMarkerGeometry = UUID.randomUUID();
             experimentalMarkerGeometryJson.put("uuid", uuidForMarkerGeometry.toString());
             experimentalMarkerGeometryJson.put("type", "SphereGeometry");
-            experimentalMarkerGeometryJson.put("radius", DEFAULT_MARKER_SIZE);
+            experimentalMarkerGeometryJson.put("radius", DEFAULT_MARKER_SIZE*experimentalMarkerScaleFactor);
```
aseth1: This still looks troublesome to me. `DEFAULT_MARKER_SIZE` does not appear to be the default user-specified marker radius from the preference, but the `currentSize`. The `experimentalMarkerScaleFactor` just scales an already scaled `currentSize` which makes the user setting meaningless. @nickbianco @chrisdembia @jimmyDunne and I all advocated for not using a `experimentalMarkerScaleFactor` because it makes sizing ambiguous. So why are we doing this?

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
```diff
@@ -1091,6 +1090,8 @@ public void setExperimentalMarkerScaleFactor(double experimentalMarkerScaleFacto
                         experimentalMarkerScaleFactor*DEFAULT_MARKER_SIZE);  
             }
         }
+        // update cached experimentalMarkerGeometryJson
+         experimentalMarkerGeometryJson.put("radius", DEFAULT_MARKER_SIZE*experimentalMarkerScaleFactor);
```
aseth1: See previous comment.

Gui/opensim/view/src/org/opensim/view/experimentaldata/MotionObjectPointForce.java
aymanhab: @aseth1 The cascading calls are used all the time to set attributes on Decorative Geometry as was done here https://github.com/opensim-org/opensim-core/blob/dc81fe58bccc3061ff26e09806ba085681a71d2d/OpenSim/Sandbox/futureMarkerPreview.cpp#L75 it was built into this API by Sherm for this reason.

getStartIndexInFileNotIncludingTime() returns the index of the data visualized by the "Arrow" in the corresponding Storage file, it gets computed and cached on construction. IndexOnBody is used for book-keeping to generate a unique name to pass to the visualizer (since these are not components with full pathnames etc.)

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
aymanhab: @aseth1 DEFAULT_MARKER_SIZE is set on construction and can only be changed by updating from the user preference, while experimentalMarkerScaleFactor is the live user specified scaleup factor with initial value of 1. I believe we agreed on a plan/resolution already regarding the handling of the scale factor at this point.

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
aymanhab: This line is actually the meat of the fix, which is that when user updates the scale factor, the cached marker geometry wasn't kept in sync, now it does. Hope this explains.

Gui/opensim/view/src/org/opensim/view/experimentaldata/MotionObjectPointForce.java
aseth1: There is no issue with the cascading functionality. The issues is with documenting the code for others to understand what it is doing. Please comment your code.

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
aseth1: Again this is an issue with variable naming and code documentation. Without expanding the full file I could see:

```
DEFAULT_MARKER_SIZE = Double.parseDouble(currentSize);
```

Up above one sees that `currentSize` is set to default. It seems this could be easily misread to be the current size in use as the variable name otherwise indicates.

aseth1: >  I believe we agreed on a plan/resolution already regarding the handling of the scale factor at this point.

My feeling is that @jenhicks and you over-ruled the team and if we are following scrum neither of you have the authority to do that.

jenhicks: Uh oh @aseth1 ... it seems you have discovered the secret cabal @aymanhab and I created to defeat the scrum [evil laugh]. Actually, I think this may just be a misunderstanding. I didn't see the timing for today's sprint on the calendar, but I am back working as usual today and happy to chat anytime via zoom.

aseth1: A simple vote will easily clarify.

aymanhab:
Comments addressed and changed functionality to use/present Marker Size (mm) instead of Marker Scale factor. Ready for review/test

nickbianco:
Tested on [AppVeyor build 1.0.1728](https://ci.appveyor.com/project/opensim-org/opensim-gui/build/1.0.1728), Windows 10. 

I am able to adjust the marker size for a previewed TRC file with absolute values in millimeters. However, the default marker size is still 5mm, not 10mm. I believe [this line](https://github.com/opensim-org/opensim-gui/blob/0eaa7e2c58cf21e0235546c09eb0706f6debcbeb/Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java#L379) may be the culprit: if the Experimental Marker Data preference is still 5mm, then I think it will override the value in the "saved" variable. 

@chrisdembia may be able to confirm this, since we looked at it together last week.

aymanhab:
@nickbianco thanks for testing. As explained somewhere else, the preferenes are persistant on Windows so if the code say default is 10 and you have 5 saved as preference then it will use 5. On osx you get fresh preferences since you can wipe old ones. You can also go the registry and remove the old Preferences on windows to simulate a fresh install as a braveheart 😄

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
```diff
@@ -1015,7 +1014,7 @@ private void createDefaultMotionObjects() {
     public static double DEFAULT_MARKER_SIZE = 5;
```
aseth1: This is confusing since the string value is "10". I would change it so the default value is defined/specified ONLY once in a single variable and then convert it to a string value for presentation.

nickbianco:
@aymanhab I didn't edit the registry, but I did try setting the preference to 10mm, then closing and reopening the GUI and that worked fine.

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
```diff
@@ -147,7 +147,7 @@
     OpenSimContext dContext; 
     ArrayStr stateNames;
     private boolean renderMuscleActivations=false;
-    private double experimentalMarkerScaleFactor;
+    private double experimentalMarkerSize=1;
```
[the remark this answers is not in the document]
aymanhab: Since the value is set in the constructor, the initialization doesn't do anything anyway, I usually initialize to non-zero value so that if it doesn't get set in constructor I can still see it (member variables initialized to 0 by default)

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
```diff
@@ -376,12 +375,12 @@ public Vec3 getDefaultForceColorVec3() {
     
     /** Creates a new instance of MotionDisplayer */
     public MotionDisplayer(Storage motionData, Model model) {
-        String saved = "5";
+        String saved = "10"; // new default per issue #643
```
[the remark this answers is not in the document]
aymanhab: I made the change @aseth1 as you suggested. Thank you 👍 Conversion from Strings to doubles and back has the potential to introduce side effects e.g "10,000001" when the property is presented to users. Current pattern keeps user entered String as is and guarantees value shows as 10. String is parsed on line 382.

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
```diff
@@ -376,12 +375,12 @@ public Vec3 getDefaultForceColorVec3() {
...
         String currentSize =TheApp.getCurrentVersionPreferences().get("Visualizer: Experimental Marker Radius (mm)", saved);
         TheApp.getCurrentVersionPreferences().put("Visualizer: Experimental Marker Radius (mm)", currentSize);
         DEFAULT_MARKER_SIZE = Double.parseDouble(currentSize);
```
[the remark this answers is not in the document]
aymanhab: - DEFAULT_MARKER_SIZE is the value set by the user as a default marker radius, I'm not sure what is incorrectly named. This value is used to initialize the member variable experimentalMarkerSize which the user can modify (without changing defaults/preferences).
 -There are no global variables in java. 
- CAPS are used traditionally to indicate constants, and if they can't be modified they're marked final. This is a constant initialized to a user preference so it's not strictly const in c++ sense. Sorry about the confusion. I removed the public modifier to make the scope clear. Thank you 👍

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
aymanhab: As explained earlier this is just an initial value that gets read from preference so it means nothing but I went ahead and changed it to 10 so it appears consistent.

aymanhab:
Changes made/accounted for in variable names/comments. Functionality can be restested if needed so the issue can move along.

nickbianco:
I tested the latest build on Windows and everything is still working fine. Moving #643 to Verified on Windows.

scrum2a:
@aseth1 finish review

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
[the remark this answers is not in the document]
aymanhab: @aseth1 using NaN in visualization is problematic since many other things can go wrong and lead to objects not shown, same as 0 which would happen if no initial value is specified. The value is set in the constructor so the initial value is irrelevant other than debugging but I made it into a 1000 so if future changes breaks the constructor, the object is still visible.

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
```diff
@@ -376,12 +375,13 @@ public Vec3 getDefaultForceColorVec3() {
...
-        String saved = "5";
-        String currentSize =TheApp.getCurrentVersionPreferences().get("Visualizer: Experimental Marker Radius (mm)", saved);
+        String defaultMarkerRadiusString = "10"; // new default per issue #643
+        String currentSize =TheApp.getCurrentVersionPreferences().get("Visualizer: Experimental Marker Radius (mm)", 
+                defaultMarkerRadiusString);
         TheApp.getCurrentVersionPreferences().put("Visualizer: Experimental Marker Radius (mm)", currentSize);
         DEFAULT_MARKER_SIZE = Double.parseDouble(currentSize);
         this.experimentalForceScaleFactor = 1.0;
-        this.experimentalMarkerScaleFactor = 1.0;
+        this.experimentalMarkerSize = DEFAULT_MARKER_SIZE;
```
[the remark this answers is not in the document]
aymanhab: It was necessary when we were doing relative scale so we know the prescale size. I went ahead and removed it. Thank you.

aymanhab:
Comments addressed, ready for merge.

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
```diff
@@ -961,7 +960,7 @@ public void updateMotionObjects(){
         if (simmMotionData instanceof AnnotatedMotion){
             // Add place hoders for markers
             AnnotatedMotion mot= (AnnotatedMotion) simmMotionData;
-            setExperimentalMarkerScaleFactor(mot.getDisplayForceScale());
```
chrisdembia: Was this a separate bug that you caught? If not, could you explain the reason for this change?

Gui/opensim/view/src/org/opensim/view/experimentaldata/ExperimentalMarkerSetNode.java
```diff
@@ -173,38 +173,38 @@ public Color getColor()
...
                public void undo() throws CannotUndoException {
                    super.undo();
-                   setMarkerScaleFactorUI(oldMarkerScaleFactor, false);
+                   setMarkerRadiusUI(oldMarkerSize, false);
                }
                @Override
                public void redo() throws CannotRedoException {
                    super.redo();
-                   setMarkerScaleFactorUI(newFactor, true);
+                   setMarkerRadiusUI(newRadius, false);
```
chrisdembia: What is the reason for changing `true` to `false`?

Gui/opensim/view/src/org/opensim/view/motions/MotionDisplayer.java
aymanhab: Yes, it was a separate bug that I caught.

Gui/opensim/view/src/org/opensim/view/experimentaldata/ExperimentalMarkerSetNode.java
aymanhab: This relates to your earlier reported bug with multiple undo/redo/undo, the change to false handles this properly by making the undo manager fully responsible for the undo/redo unwinding. Youcan test the latest artifcat and you should be able to verify. Since unrelated I can "undo" it as well.

chrisdembia: No it's fine I just mostly wanted to make sure  you intended this change.

aymanhab:
Thank you @chrisdembia 🎆

Issue #657, opened 2018-04-11
Preference to persist models results in error message on osx launch - turn off by default


jenhicks:
Since this is no longer supported

aymanhab:
Actually in my own windows environment, this works fine the two issues that I need help with are:
1. Whether it works on osx, 
2. Where to keep the AppState.xml file then
3. If this continues to work on other slow machines (when it is slow to launch server etc.)

Thoughts on OSX, @chrisdembia ? We could switch to using UserDirectory as defined by NetBeans. http://wiki.netbeans.org/FaqWhatIsUserdir

jenhicks:
As of OpenSim ed2365e-2018-04-12, persist models does not work on osx.

jimmyDunne:
Persist Model On works. Also, when switching to Off, no model is opened on reboot. 
W10. OpenSim-8d665089-2018-05-03.zip

aymanhab:
Awesome, need to try latest build on osx

jenhicks:
5-31 build on OSC: With the persist model option on, I almost always get the error message "This site can't be reached". If I click the blue "reload" button in the bottom right corner, the visualizer will appear.

PR #947, opened 2018-08-07
Make "Persist Models" default to Off

aymanhab:
Fixes issue #657

### Brief summary of changes
Default to not persist models on fresh install

### Testing I've completed
On osx with new install option is Off

### CHANGELOG.md (choose one)

- not sure if we need to update user option is still available


jenhicks:
@chrisdembia @nickbianco @jimmyDunne Could someone verify this?

chrisdembia:
I just verified this. "Persist Models" is Off by default, and the models are indeed not persisted. Merging.

Issue #483, opened 2018-01-18
Display pop-up box if an object is automatically renamed when model is opened


tkuchida:
Jen: Users should get a warning and now what steps to take if we edit their mode files.

If an object was renamed when the model was opened, the user should be notified that additional edits might be necessary in the model file (particularly if the model fails to open). Should report which objects were renamed.

aseth1:
In the API we can change warnings to Exceptions, but GUI will have catch and throw up a dialog with the message. If there are many renames (e.g. dozens to hundreds of PathPoints) this could be very agitating. Would need away to "ignore all". 

Another option, is to add/return a flag that informs us if finalizeFromProperties has changed any property values, and in that case the GUI would show a single pop-up with a message to check the messages console for the list of the components that were renamed.

aseth1:
I also would like us to avoid renaming Components that are unique by name and type if we can. Since we agreed that the pathname must be unique, we can roll the Model Sets into Components (Lists or some other Container) so that you can have: `/hopper/Bodies/foot` and `/hopper/ContactGeometry/foot` (which is pretty typical of 3.3 models) and would be easier than renaming `/hopper/foot` (the Body) and `/hopper/foot_0` (the ContactGeometry) and having to find all the contact Forces that use the `foot` ContactGeometry and rename to `/hopper/foot_0`.

The tradeoff is that the pathnames get longer (minor) but it also forces users to think they need these sets (groupings) of Components in the Model, which the Component interface does not require. We can add subcomponents to any Component. The GUI will need likely need these component Sets for the Navigator for awhile, so perhaps it is fine to use these typed sets to help differentiate components (e.g. include in pathname) since they'll be there for the foreseeable future.

jenhicks:
Add a pop-up to review the messages window.

jenhicks:
@aseth1 Any updates on this one?

aseth1:
@jenhicks There were several proposals made.  Let me summarize:
1. Change warnings into Exceptions so the GUI can notify users of renames. This will require a solution to multiple renames bogging down the user. It may also be somewhat counter-productive to throw an exception when we are doing the rename to make the transition easier for the user by avoiding the non-unique name exception. 
[...]

PR #953, opened 2018-08-14
Popup model loading errors

aymanhab:
Fixes issue #483 partially
### Brief summary of changes
If Model loading error log is not empty then popup a Warning dialog with the error messages. This is checked only on Model loading and need the log to be populated on the opensim-core side.

### Testing I've completed
Loaded multiple models and got no Popup, Will test more and check formatting when log is populated.

### CHANGELOG.md (choose one)

- no need to update because this is new functionality


aseth1:
Applied this PR to https://github.com/opensim-org/opensim-core/pull/2243. Additional changes were required to append to the validation log within `Model::finalizeProperties()` which causes the same message to be written to the messages windows (which could be OK) and appears in the console repeatedly when using the API an `finalizeFromProperties()` is invoked during model editing.

The validationLog mechanism is currently unable to present warnings or errors raised during finalizing connections or initializing the `System` and/or its `State`.

Below is the popup that appears:
https://github.com/opensim-org/opensim-core/pull/2243
image

tkuchida:
I think some users might find that message cryptic. "Please consider updating… to ensure consistent unit conversion…" is ambiguous: are units converted sometimes but not always ("ensure")? Perhaps tell them (1) what could go wrong, (2) how they would detect a malfunction, and (3) what to do if a malfunction occurs.

aseth1:
> Perhaps tell them (1) what could go wrong, (2) how they would detect a malfunction, and (3) what to do if a malfunction occurs.

(1) Coordinate values for `Coupled` coordinates can be assigned values that are in degrees if a coordinate was were previously (pre-4.0) labeled as `Rotational`
(2) This will lead to incorrect motion of bodies governed by the `Coupled` coordinate when playing back a pre-4.0 motion file, and produce erroneous velocity and acceleration estimates that yield incorrect inverse-dynamics and static-optimization results
(3) Apply the `updateKinematicsFilesForUpdatedModel(model_file, list_of_MOT_files)` utility to update motion files generated prior to 4.0, which will undo the conversion of `Coupled` coordinates to degrees so that they are consistent with the 4.0 model.

Let me know if this sounds reasonable. Please feel free to polish, reword in manner you feel will be more easily interpreted by users and I'll update the message in https://github.com/opensim-org/opensim-core/pull/2243.

aseth1:
Current message ready for review in: https://github.com/opensim-org/opensim-core/pull/2264

aseth1:
The *validationLog* log mechanism as implemented by `Model` is not sufficient to relay messages generated by `Component`s of the `Model`.  As it stands a `Component` is able to complain (throw an `Exception`) if its properties are invalid and in its `extendFinalizeFromProperties` it throws or handles them but there is not mechanism to write to a common *validationLog* and certainly no way to get access to the Model's log.

In the case of duplicates they are renamed automatically and we can write a message to the console, but the Model's validation log is not accessible by the `Component`. `Component` is not and should not be aware of concrete types so I rejected the idea of traversing up the ownership tree to get the `Model` to append to its log. 

Pushing *validationLog* down to `Component` is possible but then the issue is handling the fact that every `Component` has a log and composing them is not trivial (and a potential waste).  We could make *validationLog* a singleton shared by all `Component`s. Then the same log can be accessed by different components, which may not be desirable. 

@aymanhab @chrisdembia @tkuchida how would you like the messaging to work (at first)? @aymanhab i think your intention was to keep it very simple and have a direct way for Model to relay warnings to users during loading, but it seems to me we might want to control what bubbles up to users and what doesn't based on a warning/messaging level, and in that case it seems what is in place will not work.

This reminds me of discussions in https://github.com/opensim-org/opensim-core/issues/36. Perhaps we can leverage an existing mechanism.

aymanhab:
@aseth1 While the hierarchical model/component layout is great for distributing responsibility for development purposes, the model is the most common object that reads from file from a user persepctive and so it makes sense for it to be the only object responsible for reporting file parsing/handling-old-format errors, that's why the validationLog lives with the model. 

Having individual components throw errors/excpetions to the outside world would be problematic because in general components maybe too low level to know the context. if they throw exceptions I think these need to be caught and reported by the model  (in the validationLog or whatever final mechanism we decide on). It may be possible when reading from old file format to do two passes, the first one is what we do today, if that results in any errors/exceptions, a check/validate pass is performed  that returns detailed errors/warnings. the validation log could be passed along from the model for that second pass.

aymanhab:
@aseth1 I updated the branch to show html capable dialog now, you can give it a spin and let me know how it goes. Although you can't follow the links the user will be able to copy the URLs and paste into the browser, also some minor formatting support.

aseth1:
I don't think the message window has improved with the latest change:

image

Newlines are ignored so spacing is not great, looks like an editable form and it also remains editable.

aymanhab:
@aseth1 Sure I can help make this better. Keep in mind that since this is html now that new lines are ignored and the formatting would need to use <br> instead but you get to use other html as well for bold etc. if it helps.

aseth1:
@aymanhab I think I prefer the look (not the text) of the previous popup error window. We can ask @jenhicks @tkuchida @chrisdembia  for their preference.

scrum2a:
In scrum, we agreed to go back to original warning dialog.

aymanhab:
Last commit reverted, should have the standard warning dialog back at head of branch @aseth1

chrisdembia:
> @aymanhab @chrisdembia @tkuchida how would you like the messaging to work (at first)?

I'm fine with each component having a *validationLog*.

chrisdembia:
I'm not merging this yet because of @aseth1's discussion about the *validationLog* mechanism.

aymanhab:
@chrisdembia As of now validationLog lives with the model because it is intended to capture model loading errors. One problem with having it live with Component is that it would live long post construction which is unnecessary, and even for programmatically created components. It may actually make sense to tie it to the XMLDocument to limit this lifetime, and pass it along to parsing and post-parsing validation.
- It's important that in the most common case we have a simple test to check for errors/warnings in one place rather than try to assemble a long list of mostly empty error logs to figure out if something went wrong.

chrisdembia:
Those are good points. We should see what other software packages do. I don't think we should address the bigger validationLog issue now though.

scrum2a:
@aseth1 please respond about `validationLog`.

aymanhab:
Based on discussion with @aseth1 at Sprint planning I think we agreed this can be merged and the discussion moved to separate issue to address a more general mechanism for communicating errors/warnings at component level. Can you please confirm, @chrisdembia Thanks

Issue #736, opened 2018-05-11
GUI Scripts need update: test all and fix or remove from distribution


aymanhab:
Jen: Users need examples to show key GUI scripting functionality and at minimum any examples we distribute should work without bugs

### Steps to reproduce
Try to run runtutorialOne.py GUI script fails because the models are not in install folder anymore.

### Expected result
tutorial runs
### Actual result
Exception/error due to not being able to locate model.

### Environment and GUI version
Any

aymanhab:
@chrisdembia The scripts used getInstallDir() which was built into the code, now we need to use the Resources directory instead.  I can add a function to get it since we keep it as a Preference. Any thoughts? Do you think we should have the preference exposed?

chrisdembia:
I think it would be great if the GUI maintained the last location the user specified to install the resources (the preference gets updated), and the user could also view and manually edit this location through the preferences. And yes, it would be great to access this directory within the scripting shell (`getResourcesDir()` or whatever name is consistent with other preferences/functions). I hope this isn't inconsistent with other suggestions I've given!

I thought the preference was already exposed?

aymanhab:
@jimmyDunne can you help update the scripts and report what works and what doesn't so we can either fix or exclude when you have a chance? Thanks

aymanhab:
@jimmyDunne The method  getResourcesDir() has been introduced in commit dated 5/11/18

* Expose method to scripting shell (getResourcesDir) so that GUI scripts can use it to locate models and data files

Please let me know if you run into issues updating the GUI scripts to use it. Thanks

jimmyDunne:
This is taking a little longer than expected due to changes in the API. For example, the code that changes the path points in the wrist model have significantly changed since 3.3.  I am getting through the scripts, but it turned out it isn't so simple. 😭 

On the brighter side, getResourcesDir() works great! Thanks, Ayman!

jenhicks:
@jimmyDunne Let's focus on testing (the sprint tasks) before making big changes to these scripts.

jimmyDunne:
GUI Scripts
- [ ] alterTendonSlackLength
- [ ] createControlsFromStorage
- [ ] editProperties
- [ ] makeUlnaHeavy
- [ ] modelBuildingUtilities
- [ ] plotMuscleFiberLengthAgainstFile
- [ ] plotMuscleMomentArmAgainstMotion
- [ ] runMultipleIKTrials
- [ ] runScaling
[...]

Issue #941, opened 2018-08-06
Always installResources() on a fresh install


jenhicks:

aymanhab:
@chrisdembia Please let me know if you need help working on this one or I can take it over. Thank you

aymanhab:
@chrisdembia I think we will need a build-id and have it accessible from the GUI code so that we can check for a new build and use that to call installResources. This will be very helpful anyway for frequent build/distributions long term maintenance.

chrisdembia:
Should we use a combination of the opensim-gui, opensim-models, and opensim-core commit hashes?

aymanhab:
But opensim-models is a submodule, it needs new commit to make it to opensim-gui
HHashes are good but don't provide order so date followed by gui-commit may be good enough ID

chrisdembia:
Ah good point. However, `opensim-core` is not being used as a submodule (even though it currently exists as a submodule).

I'm fine with adding the date as well.

aymanhab:
I wasn't prompted to install resources on subsequent builds so that part seem to have worked as expected.

aymanhab:
If anybody has cycles to verify this that would be great. @carmichaelong @jimmyDunne @nickbianco Specifically:
- install build from PR
- Verify you get prompted to installResources on first launch and never after
- Report what you see in About dialog for "Product Version"
I got expected results on my windows machine but would like to see what others see as well.

chrisdembia:
I can fix the CI scripts to avoid the version appearing twice.

aymanhab:
Any volunteers to verify on either Mac or Windows @jenhicks @carmichaelong @nickbianco @jimmyDunne

jenhicks:
I tried out on Mac. I think everything worked as expected. See images below.

![installresources](https://user-images.githubusercontent.com/4510959/44550751-fc8ccc80-a6d9-11e8-9eb8-8276220b0608.png)
![about](https://user-images.githubusercontent.com/4510959/44550756-ff87bd00-a6d9-11e8-8fdb-e74dfa3cc45a.jpeg)

aymanhab:
Perfect, thank you @jenhicks

chrisdembia:
Verified on Windows 4.0-2018-08-23-480caf33.

PR #958, opened 2018-08-15
update "OpenSimResourcesDir" preference on successful install

aymanhab:
so that future calls from Scripts find the latest installed Scripts/models

Fixes issue #736 partially

### Brief summary of changes
Made installResources() internally updates OpenSimResourcesDir(). This may become less relevant when we installResources on every fresh install (issue #941)

### CHANGELOG.md (choose one)

- updated...


aymanhab:
Thanks @chrisdembia for the amazing quick response! 🥇

Issue #346, opened 2017-07-28
Analytical shape edits are not propagated to visualizer


aseth1:
The `Cone`'s *direction* property has no effect on its orientation. Always points upward (along +Y), regardless.

aymanhab:
No examples yet, should be easy to fix once we have a model to test with.

chrisdembia:
Here you go: [cone.osim.txt](https://github.com/opensim-org/opensim-gui/files/1876920/cone.osim.txt) (remove .txt extension).

The cone height and cone direction are not obeyed. I think this is an opensim-core issue.

jimmyDunne:
Changing the cone height or radius had no effect. 

![cone_model.gif](https://images.zenhubusercontent.com/544fc8297c6ccc6277a5b651/b36a5e66-a508-4f83-bc3a-3da0e3f4bcaa)

jimmyDunne:
This is issue is to do specifically with the Cone. If there are filed issues (and PR's) elsewhere for the different shapes, it would helpful for you to link them here.

aymanhab:
PR #871 hides parameters that are not handled yet (needs fix on API side for orientation, similar to cylinders)

chrisdembia:
1. **Brick**: I am able to edit the `half_lengths` of a `Brick` ✔️ , and the visualizer updates appropriately, and the quadrants property does not appear ✔️ .
2. **Ellipsoid**: (a) Editing the `radii` for an`Ellipsoid` has no effect ❌ . (b) The quadrants property is not hidden. ❌ 
3. **Sphere**: Editing `radius` ✔️ ; `quadrants` property is not hidden ❌ .
4. **Cylinder**: Editing `radius` ✔️ and `half_height` ✔️ ; `quadrants` property is not hidden ❌ .
5. **Cone**: Editing `height` ✔️ and `base_radius` ✔️ ; `quadrants` property is hidden ✔️ .

Why was this moved to "Needs Verifying"?

aymanhab:
@chrisdembia In the past no changes to Analytical shapes were ever propagated. I agree we should either support or hide quadrants, looking for feedback on that. Elipsoid is implemented as a scaled Sphere, we can hide radii and users modify scale factors. Thoughts?

aseth1:
What do `quadrants` mean in the context of `Geometry`? If it is only necessary for wrapping then `WrapGeometry` should define it.

chrisdembia:
[...]

PR #965, opened 2018-08-20
No need to hide quadrants since they were removed from opensim-core

aymanhab:
Fixes issue #346

### Brief summary of changes
Remove the line that hides quadrants Property from PropertyEditor in the GUI since Property "quadrants" has been removed from "AnalyticGeometry" types.

### Testing I've completed
Built gui and verified all the checkmarks tested earlier by @carmichaelong still check and that no quadrants property is shown in the GUI

### CHANGELOG.md (choose one)

- no need to update because Geometry is a new feature in 4.0


aymanhab:
Thank you @aseth1 much appreciated.

Issue #303, opened 2017-07-17
Adding/Deleting PathPoint of GeometryPath fails when it has wrapping surfaces


aseth1:
Using the GeometryPath editor (via Properties editor for a Muscle's GeometryPath) adding or removing a PathPoint causes exceptions. Requires saving the edited model, quitting the GUI, deleting the JXBrowswser cache, relaunching GUI and reloading to see the change.

Exception:
```
java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Long
	at org.opensim.threejs.JSONMessageHandler.convertLocationStringToPropertyFormat(JSONMessageHandler.java:71)
	at org.opensim.threejs.JSONMessageHandler$1.run(JSONMessageHandler.java:37)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:76)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:726)
	at org.netbeans.core.TimableEventQueue.dispatchEvent(TimableEventQueue.java:159)
[catch] at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:109)
	at java.awt.WaitDispatchSupport$2.run(WaitDispatchSupport.java:184)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:229)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:227)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.awt.WaitDispatchSupport.enter(WaitDispatchSupport.java:227)
	at java.awt.Dialog.show(Dialog.java:1084)
	at org.netbeans.core.windows.services.NbPresenter.superShow(NbPresenter.java:1086)
[...]

PR #967, opened 2018-08-23
Edit geometrypath with wrapping

aymanhab:
Fixes issue #303 
### Brief summary of changes
On GeometryPath edit, recreate visualization completely rather than incrementally insert/remove pathpoint visualization artifacts.

### Testing I've completed
Add/Delete/Change-frame of path points all work as expected. With so many possible pathways to edit (add/delete/change-frame at end-points and in the middle of the path) it would be good to have another pair of eyes test the change to make sure all scenarios are handled properly.

### CHANGELOG.md (choose one)


aymanhab:
Need opensim-core PR https://github.com/opensim-org/opensim-core/pull/2268 so that inserting a new pathpoint places it in sensible location.

PR #970, opened 2018-08-27
Update files created by running SWIG on latest opensim-core

aymanhab:
Fixes issue #0

### Brief summary of changes
This PR upgrades SWIG output (after inspection) to be in sync with opensim-core API

### Testing I've completed
Built GUI and launched it.

### CHANGELOG.md (choose one)

- no need to update because this is internal change


aymanhab:
Thanks @chrisdembia 👍

Issue #949, opened 2018-08-11
External Loads Pane exception thrown when .xml is version=2


jimmyDunne:
This seems to be a problem between GUI and the API in what constructor to call...

If you use the STOFileAdapter() to write forces to a .mot file, the motion file will have version=2. When I try to use the written mot file with version=2 in the GUI (External Loads Panel)I get an exception from Storage(). 

Message window shows 

> Storage: FileAdpater failed to read data file.
> Cannot read headers only if not a STO file or its version is greater than 1.
> 	Thrown at Storage.cpp:206 in Storage().

The section causing the caught exception in storage seems to be [here](https://github.com/opensim-org/opensim-core/blob/dae70d09e419eac1b37d16a193f53b526a513135/OpenSim/Common/Storage.cpp#L201-L224). 

I don't get this issue in Matlab API so I checked the GUI and it seems like the GUI may be [using a constructor](https://github.com/opensim-org/opensim-gui/blob/71c8c385873b86900d6c46aaf8595227a33b4d44/Gui/opensim/tracking/src/org/opensim/tracking/EditExternalLoadsPanel.java#L134) that is tripping the exception.

This may be an API issue, a GUI issue, or both— not sure.

chrisdembia:
@aymanhab what is the reason that the GUI passes `readHeadersOnly=true`?

aymanhab:
@chrisdembia The GUI doesn't actually use anything other than the headers to establish mapping, no need to read/parse a potentially huge file unnecessarily to throw it away.

chrisdembia:
Thanks @aymanhab then we need to chat about how to fix this in opensim-core.

chrisdembia:
@aymanhab have you already started working on this? I could look into it.

aymanhab:
Not yet but I can work on it today. Let me know if there's anything else that came up in scrum that's higher priority

chrisdembia:
Nothing else came up.

chrisdembia:
Okay I will work on other things then.

aymanhab:
How do I get a version 2 .mot file @chrisdembia @jimmyDunne

chrisdembia:
@aymanhab I created the attached file using the `STOFileAdapter`.

[subject01_walk_grf_version2.mot.txt](https://github.com/opensim-org/opensim-gui/files/2329891/subject01_walk_grf_version2.mot.txt)

aymanhab:
Thanks @chrisdembia

[...]

PR #974, opened 2018-08-28
Read whole data file to define external loads 

aymanhab:
not just header to workaround failure to handle read-headers-only in mot/sto version >= 2

Fixes issue #949

### Brief summary of changes
Used constructor of Storage that reads whole file rather than headers only.  

### Testing I've completed
Checked there's no other reference to the constructor that utilizes read-headers-only. Was able to define ExternalLoads without issue after building the GUI. 

### CHANGELOG.md (choose one)

- no need to update because this is internal.


aymanhab:
@chrisdembia I tested with the file you sent me (v2) and was able to create an externalloads file from the dialog. Other mot files in repo/examples are all version 1. What do you think? Actual mot files v2 result from c3d files which I don't know if we have a working example of anyway. Thoughts?

chrisdembia:
@aymanhab was there a noticeable delay when selecting the MOT file?

aymanhab:
Not really,  rather unnoticeable

aymanhab:
Thank you @chrisdembia

PR #976, opened 2018-08-29
Remove motion displayer assemble

aymanhab:
Fixes issue #0

### Brief summary of changes
Remove call to model.assemble when playing back results from file (after calling formStateStorage)

### Testing I've completed
Bouncing-block FD and playback works fine and doesn't violate contacts/constraints

### CHANGELOG.md (choose one)


aymanhab:
Note that leg69 still shows patella totally off  without assemble but this is due to opensim-core issue https://github.com/opensim-org/opensim-core/issues/2240

aymanhab:
@chrisdembia @aseth1 it would help test issue https://github.com/opensim-org/opensim-models/issues/87 if this was merged. Thoughts?

chrisdembia:
Well now that you've created the PR, we have a build with which to test https://github.com/opensim-org/opensim-models/issues/87. Which should be merged first, this PR or https://github.com/opensim-org/opensim-models/issues/87?

I think we should merge this PR no matter what, because leaving in the assemble can cause erroneous playback.

aymanhab:
I agree, and it will make it clear if more models/motions need motionType fix.

chrisdembia:
Perhaps we can discuss with @aseth1 in the stand-up.

aymanhab:
@chrisdembia I think we can merge this based on the verbal agreement in scrum meeting, will wait for your formal review/approval.

aymanhab:
Thank you @chrisdembia 👍

PR #977, opened 2018-08-29
Fix preference initialization in installer.java

aymanhab:
Fixes issue #643 

### Brief summary of changes
Default Experimental Marker radius is specified in two places, one of them was not updated. Fixed in this PR and added comment to indicate that change in one place should update the other.  

### CHANGELOG.md (choose one)


aymanhab:
A more general but significant change would be to create a singlteon DefaultPreferences object to hold values, eventually use that to reset Preferences to Defaults and use it across the board to initialize all preferences. Which would address issue #961 as well

Issue #955, opened 2018-08-15
The Scripts preference seems to have no effect


chrisdembia:
There's a Scripts Dir preference but it has no effect on the Scripts menu on Mac.

![](https://user-images.githubusercontent.com/846001/44113233-3f0d75d4-9fbc-11e8-9f0d-ccf9ca7949b1.png)

Raised in #917.

chrisdembia:
The behavior of the `getScriptsDir()` command and of the Scripts preference are documented on Confluence:

- https://simtk-confluence.stanford.edu:8443/display/OpenSim/Common+Scripting+Commands
- https://simtk-confluence.stanford.edu:8443/display/OpenSim/Scripting+in+the+GUI

> You can change the location of the Scripts directory by going to Edit > Preferences and changing the directory under the "ScriptsPath" option (see User Preferences for more information). 
Directories are relative to the OpenSim installation Directory (e.g. Specifying "MyScripts" will look for scripts in the folder C:\ProgramFiles\OpenSim3.0\MyScripts if you installed OpenSim in the folder C:\ProgramFiles\OpenSim3.0). 

I suggest the following:

- make `getScriptsDir()` and `Paths: Scripts Path` consistent (dir vs path),
- change the preference to be an absolute path, and
- the default value of the preference should be `<ResourcesDir>/Code/GUI`.

PR #981, opened 2018-08-30
Default scriptsPath to ResourcesDir/Code/GUI

aymanhab:
Also populate cascade menu from it as described in user guide

Fixes issue #955

### Brief summary of changes
getScriptsDir() was renamed to getScriptsPath() and wired to be initialized to the new location of Scripts <Resources>/Code/GUI

### Testing I've completed
Made installer locally and behaved as expected/described in User/Guide 
### CHANGELOG.md (choose one)


chrisdembia:
On Mac, `Scripts Path` is `null/Code/GUI/`

image

aymanhab:
Thanks @chrisdembia it's likely an initialization order issue where the preference is set before the OpenSimResourcesDir is populated, which happens actually after the user selects a directory to copy resources. I'll look into this.

aymanhab:
Issue addressed by moving initial setting along with InstallResources, this artifact should be good to test all the latest scripts merged by @jimmyDunne once build is ready.

jimmyDunne:
Works great! Gives the correct path to scripts!

chrisdembia:
@aymanhab what is the reason for changing the `Scripts` menu?

before:
image

after:
image

chrisdembia:
The new menu organization seems redundant.

Also, on Mac, I am not getting a `Scripts Path` preference, even after running `defaults delete org.opensim.version40`, running `installResources()` and restarting the GUI:

image

aymanhab:
Thanks for the catch @chrisdembia I missed the change in the menu and apparently @jimmyDunne didn't see it either, will revert the menu change to keep 3.3 layout.
The preference should show up if you allow the installer to copy the resources for you, not sure what happens if you cancel out since preferences on windows and osx are different. Once you see it you should be able to verify it's working as documented. Please let me know otherwise.

aymanhab:
Sorry, I mean on first launch. You can delete the file we use to mark first
launch manually.

Issue #975, opened 2018-08-28
Remove all references to vtk in GUI code and remove vtk module


aymanhab:
### Actual result

(include screen grabs where appropriate)

PR #983, opened 2018-08-31
Remove vtk

aymanhab:
Fixes issue #975 

### Brief summary of changes
Remove dependencies on vtk (already broken functionality since libraries aren't distributed any more) and remove vtk module altogether to cleanup code base and simplify code maintenance.

### Testing I've completed
Built GUI and launched it.

### CHANGELOG.md (choose one)

- no need to update because cleanup


aymanhab:
Can anybody try this PR branch to see if something broke? This PR removes all traces of vtk, remove dependency and makes the distribution smaller at the same time. Also makes it easier to search the code. Whoever has cycles to donate would be great: @chrisdembia @carmichaelong @nickbianco @jimmyDunne Thank you.

carmichaelong:
I ran the three tutorial GUI scripts. I also ran the standard pipeline of Scale --> IK --> RRA --> CMC using gait2392 and the accompanying files. These all seemed to work fine.

I ran into one problem where I (incorrectly) tried to run IK on a model without markers then tried to open the plotter. I tested this on the master branch and the problem was also there, so the issue is not with this branch (#1004 details this).

aymanhab:
Thanks much @carmichaelong 🥇

PR #987, opened 2018-09-04
Restore scripts menu to standard layout

aymanhab:
Fixes issue #955 

### Brief summary of changes
PR #981 had a side effect of changing the Scripts menu. This PR undoes this change and also sets the ScriptsPath preference when user forces installResources()

### Testing I've completed
Tested on Mac (where Preferences are not persistent) and all worked as expected
### CHANGELOG.md (choose one)


aymanhab:
@chrisdembia this PR undoes the unintentional change to Scripts menu. Feel free to test the Artifact from the build as needed.

Issue #992, opened 2018-09-07
Slow refresh of Path display upon structural edits or Restore


aymanhab:
### Steps to reproduce
Add, remove or change type of a PathPoint then do Restore, the path takes up to few seconds to redraw.

### Expected result
instant redraw
### Actual result
takes up to few seconds to redraw
(include screen grabs where appropriate)

aymanhab:
I'm investigating a fix, if it turns out to be limited scope fix then I'd make it into 4.0, otherwise post 4.0. I understand and support that we don't make far reaching changes at this point.

PR #995, opened 2018-09-11
Update threejs module

aymanhab:
Fixes issue #992

### Brief summary of changes
Avoid expensive addition of nameless intermediate points on Path to scene as this triggers sceneGraph modification sequence.

### Testing I've completed
Faster redraw of paths on restore

### CHANGELOG.md (choose one)

- updated...


aymanhab:
Folded into a later PR

PR #999, opened 2018-09-14
modelBuilder utility class to add Bodies, Joints and Geometry

aymanhab:
Fixes issue https://github.com/opensim-org/opensim-core/issues/2248

### Brief summary of changes
New class with static methods so inline with typical workflow that allows addition of Body, Joint and Geometry

### Testing I've completed
Ran script at bottom of file in GUI and worked as expected and was able to modify the Geometry/transform

### CHANGELOG.md (choose one)

- updated...


aymanhab:
@chrisdembia this builds on @aseth1 model_building_utilties. Needs more error checking. Are there better ways to instantiate Joints? Feedback is appreciated. @carmichaelong

chrisdembia:
How does this relate to the existing `modelBuildingUtilities.py` script in the repository?

Gui/opensim/Scripts/modelBuilder.py
```diff
@@ -0,0 +1,45 @@
...
+	"""Class to contain model building utilities."""
+	@staticmethod
+	# jointType is one of: 
+	# 'PinJoint', 'FreeJoint', 'WeldJoint', 'PlanarJoint', 'SliderJoint', 'UniversalJoint'
+	def connectBodyWithJoint(model, bodyName, jointName, jointType):
+		c = model.getComponent(bodyName)
+		c = modeling.PhysicalFrame.safeDownCast(c)
+		p = model.getGround()
+		module = sys.modules['org.opensim.modeling']
+		JointClass = getattr(module, jointType)
```
chrisdembia: I think you could just directly use `org.opensim.modeling` as the first argument.

chrisdembia:
> Are there better ways to instantiate Joints?

Your method for instantiating joints looks good to me.

aymanhab:
This is ready for review @carmichaelong @chrisdembia @jimmyDunne

carmichaelong:
LGTM. I think this could be merged as is.

Only comment is to perhaps add `CustomJoint` as a separate case in `connectBodyWithJoint()` and return some message about: 1) What to do now if you want to add a `CustomJoint` and/or 2) mention that it will be supported in the future. If this seems out of scope, feel free to merge.

aymanhab:
Thanks @chrisdembia I'll merge once we get @carmichaelong or @nickbianco to try it since they were in the discussion regarding model buildingso want to make sure we met their expectations

aymanhab:
Thanks @carmichaelong Will merge and address CustomJoint separately.

Issue #1013, opened 2018-10-12
Mac app icon is gone, and name of app is lowercase "opensim"


scrum2a:
@aymanhab may have inadvertently edited an Ant/Netbeans config file.

aymanhab:
Thanks @chrisdembia I didn't download the artifact and run on osx, I assume you did and it worked. Thanks again 🥇

chrisdembia:
Yes I did.

PR #1017, opened 2018-10-12
Restore branding that was lost by earlier commit

aymanhab:
Fixes issue #1013 

### Brief summary of changes
Restore branding/application name that was modified by commit removing dead code.
### Testing I've completed
None, will test on OSX artifact when PR is built.
### CHANGELOG.md (choose one)


chrisdembia:
I tested this and it works. 

Fixes #1013.

Issue #1008, opened 2018-10-01
Test common GUI workflows


scrum2a:
We have made a bunch of changes to core and GUI since the last time we did user testing of the GUI. We should use the GUI artifact for the sets change. Download it by clicking "Details" next to "appveyor/pr" here: https://github.com/opensim-org/opensim-gui/pull/1005

- [x] tutorials (5 pts, Carmichael successfully tested Tug of War with OpenSim 4.0-2018-10-01-76939822)
- [x] tools that use ExternalLoads (5 pts)
- [x] MarkerSets (loading/saving markers from files) (5 pts Ayman)
- [x] loading RRA adjusted model crashes due to lingering `ExternalLoads` in model's `ComponentSet` (#1005) [subject01_simbody_adjusted.osim](https://github.com/opensim-org/opensim-gui/files/2426285subject01_simbody_adjusted.osim.txt)
- [x] Creating an ExternalLoads from scratch in the GUI (8 pts) (created ExternalLoads for inverse dynamics with OpenSim 4.0-2018-10-01-76939822)
- [x] Double clicking Bodies, Muscles, ... nodes throws exception (likely because sets created on the fly for GUI use are not wired correctly to underlying model (they don't own their objects either)
- [x] MarkerSet constructor from file still has wrong signature, doesn't throw if it fails.
- [x] Deleting Markers does not work because it deletes from set not component tree

aymanhab:
Trying to load leg6dof9musc model included with build, got message about knee_angle_pat_r wrong coordinate type. Not sure if this is specific to this build or master version of model is out of date as well.

aymanhab:
@aseth1 @jimmyDunne I'm looking into MarkerSet issues now

aseth1:
> Trying to load leg6dof9musc model included with build, got message about knee_angle_pat_r wrong coordinate type. Not sure if this is specific to this build or master version of model is out of date as well.

leg6dof9musc from models repo loads without error, but the one distributed with the build does.  Seems specific to the build.

aymanhab:
Saving works fine but loading saved markers fails and throws exception

aseth1:
> Double clicking Bodies, Muscles, ... nodes throws exception (likely because sets created on the fly for GUI use are not wired correctly to underlying model (they don't own their objects either)

[...]

PR #1023, opened 2018-10-15
Use model interfaces to add/remove markers rather than marker set

aymanhab:
Fixes issue #1008

### Brief summary of changes
Instead of performing edit operations on MarkerSet, addition of Markers, as well as loading from files now go through the Model class interfaces (addMarker and updateMarkerSet) after done with changes, full initSystem is invoked, enforcing that component traversal remains intact and that Component ownership tree is up-to-date.

### Testing I've completed
Loaded a model with markers, saved the markers to a file, deleted the markers form the model and then loaded the markers from the file, model files were identical before and after. Also tested loading a markerset file that contains names already in the model. In these cases, error is given and the new markers are ignored.

### CHANGELOG.md (choose one)

- no need to update because this is a bug fix

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/markerEditor/NewMarkerAction.java
+++ b/Gui/opensim/view/src/org/opensim/view/markerEditor/NewMarkerAction.java
@@ -77,11 +77,8 @@ public class NewMarkerAction extends AbstractAction {
         Model model = markersNode.getModelForNode();
         Vec3 offset = new Vec3(0.11, 0.22, 0.33);
         MarkerSet markerset = model.getMarkerSet();
-        PhysicalFrame frameToUse;
-        if ( model.getBodySet().getSize()==0)
-            frameToUse = model.get_ground(); // model only has Ground
-        else// Model has at least one body in BodySet.
-            frameToUse = model.getBodyList().begin().__deref__();
+        // Alays create new marker in Ground, user can change that later
+        PhysicalFrame frameToUse = model.get_ground(); 
         String newMarkerName = makeUniqueMarkerName(markerset);
         Marker newMarker = new Marker();
         newMarker.setName(newMarkerName);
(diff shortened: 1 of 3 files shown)


Gui/opensim/view/src/org/opensim/view/markerEditor/NewMarkerAction.java
```diff
@@ -77,18 +77,19 @@ public void actionPerformed(ActionEvent evt) {
...
         MarkerSet markerset = model.getMarkerSet();
-        Body body = model.getBodySet().get(0);
-        if (body == null) {
-            return;
-        }
+        PhysicalFrame frameToUse;
+        if ( model.getBodySet().getSize()==0)
+            frameToUse = model.get_ground(); // model only has Ground
+        else// Model has at least one body in BodySet.
+            frameToUse = model.getBodyList().begin().__deref__();
```
chrisdembia: Why not always use ground?

aymanhab: Maintain old behavior but possible to always use Ground

chrisdembia: Ground will not be in the BodySet, so I think this actually changes behavior.

aymanhab: The old code would actually crash if there are no Bodies in the model altogether, so this was a defensive change, but I agree that creating the new Marker in ground always would be more consistent with old behavior. Will do :+1

aymanhab:
@carmichaelong If you have cycles you can check that the functionality works using the artifact in terms of save_to_file/restore_from_file and add/delete Markers.

carmichaelong:
@aymanhab I tested the following:
1) deleting roughly half the markers from leg39
2) saving markers to file
3) deleting the remaining markers
4) loading markers to file (successfully getting back to step 2)
5) adding a marker

I had tangential issues that I opened up in #1025 and #1026. If you believe that those are not related to this PR, then this PR LGTM on Windows.

aymanhab:
Thanks @carmichaelong for testing, the issues reported are unrelated to this PR so will merge once @chrisdembia approves. I will look into the issues you reported separately.

aymanhab:
Thanks @chrisdembia 👍

Issue #1025, opened 2018-10-16
Cannot change a Marker's parent_frame using Socket drop-down menu


carmichaelong:
1. Open leg39
2. Navigate to any marker
3. Try to change the `PhysicalFrame:parent_frame` Socket to anything else
4. Exception is thrown that the Component can't be found:
![image](https://user-images.githubusercontent.com/7455790/47052730-813a1b00-d15e-11e8-9020-019d03cc324a.png)

W10 Pro, OpenSim 4.0-2018-10-16-6011b19b

Issue #1018, opened 2018-10-15
Child/parent frame sockets cannot disambiguate between two frames with the same name


carmichaelong:
Testing was done using arm26.osim on W10 Pro with OpenSim 4.0-2018-10-12-3071a2d5:

Now with the Joint upgrade code always adding two offset frames, the pull-down for child/parent frame for Sockets has two frames with the same name (note both `base_offset` and `r_humerus_offset` below):
![socket-same-names](https://user-images.githubusercontent.com/7455790/46970126-0f35d900-d06d-11e8-9248-2c6019c6c306.png)

There are two types of problems when an error pops up involving one of these duplicate names:

**Case 1**
1. Open arm26
2. Navigate to Joints --> offset
3. Change `PhysicalFrame:child_frame` to `r_humerus_offset` (either one)
4. Error thrown because there were multiple frames with the same name:
![image](https://user-images.githubusercontent.com/7455790/46970276-866b6d00-d06d-11e8-86e7-5c7960a48d91.png)

**Case 2**
1. Open arm26
2. Navigate to Joints --> offset
3. Change `PhysicalFrame:child_frame` to `base_offset` (either one)
4. Although there are two `PhysicalOffsetFrame`s with the name `base_offset` (the other is owned by Joint `r_shoulder`), either one will only choose the `base_offset` owned by the Joint `offset`. No error/warning is shown to the user.

@aymanhab

aymanhab:
This should address issues #1018, #1025, leftover of #825

Issue #825, opened 2018-06-02
Navigator shows two offset frames with the same name


tkuchida:
### Steps to reproduce
- open Rajagopal2015 model
- in Navigator, expand Bodies/femur_r

### Expected result
Each Component has a unique absolute path and, therefore, a unique name on each level of the tree.

### Actual result
There are two items named "femur_r_offset" under the "femur_r" Body. The parent_frame of the patellofemoral_r Joint is "femur_r_offset", but which of the two "femur_r_offset" frames is not clear.

![](https://user-images.githubusercontent.com/4203505/40869822-a28c94c0-65d6-11e8-9426-24682942915b.png)

![](https://user-images.githubusercontent.com/4203505/40869809-3e34d19a-65d6-11e8-9b02-354ead8cd1df.png)

### Environment and GUI version
AppVeyor artifact OpenSim-cedd8d22-2018-05-31-win64.exe

aymanhab:
This boils down to how to allow editing joints. If we go down the ownership tree then the frames are not shown but joints can't be moved. Totally open to suggestions and the change is quite easy. As of now we go down from the top level model find all Frames whose BaseFrame is the Body and create frame nodes for them. In general, as long as we try to map the hierarchical ownership tree to flat list and not show full path names we'll likely have duplicates.

tkuchida:
Anything would be an improvement over showing two identical items in the drop-down list.

aymanhab:
Ok, may make more sense to move the frames under the joint. Will try this
out.

aseth1:
Why do offset frames have to be "moved" anywhere? If you just display under the `Component` (`Body` or `Joint`) they belong to, doesn't this problem go away?

tkuchida:
> If you just display under the Component (Body or Joint) they belong to, doesn't this problem go away?

Only in the first screencap. Shown in the second screencap is the (single) drop-down box for specifying the `parent_frame` Socket by name. Uniqueness is guaranteed only for absolute paths, so either the drop-down box would need to be populated with absolute paths (which might be long), or there would need to be a separate field for specifying, e.g., the base Body (essentially the first part of the absolute path).

aseth1:
Popup lists should never be populated with identically named items - it should be an error case. The pop-up list could be populated with just component names if they are all unique, otherwise it could start working up the component's path until uniqueness is achieved between components with the same name.  There has to be strategy that works because we know the component paths are unique.

tkuchida:
[...]

PR #1027, opened 2018-10-17
Drop downs for Frames will show full path names 

aymanhab:
This will help disambiguate names that could be ambiguous

Fixes issue #1025

### Brief summary of changes
When populating lists of Frames, use AbsolutePathString rather than getName() this list is used in all places where Frames are used for selection

### Testing I've completed
Changed frames for Marker(s) without an issue

### CHANGELOG.md (choose one)

- May need to update because in the past we showed only names and they were short and unique


carmichaelong:
LGTM. Tested on W10 Pro OpenSim 4.0-2018-10-17-a1ed8256.

image

Does this also address #1018?

aymanhab:
This should address issues #1018, #1025, leftover of #825

aymanhab:
Thanks @chrisdembia

Issue #1026, opened 2018-10-16
Drop-down menu for a Marker's parent_frame is populated with the current Model's Frames rather than its own Frames


carmichaelong:
1. Open leg39
2. Open arm26
3. Navigate to leg39 -> R.ASIS
4. Open drop-down menu for Socket --> `PhysicalFrame:parent_frame` and note that the `Frame`s for arm26 are shown rather than leg39
![image](https://user-images.githubusercontent.com/7455790/47052828-f3aafb00-d15e-11e8-878f-04a74d4dfedd.png)

W10 Pro OpenSim 4.0-2018-10-16-6011b19b

aymanhab:
This actually occurs in 3.3 as well, still a bad bug 🔥

PR #1037, opened 2018-10-24
When showing list of Frames in dropdown, use correct model 

aymanhab:
Pick model that the node belongs to rather than current model to create list of available frames

Fixes issue #1026 

### Brief summary of changes
FrameEditor now adjusts to the model of object being edited rather than assume current model regardless
### Testing I've completed
Loaded 2 models, tried editing joint frames, got correct frame selection/dropdown

### CHANGELOG.md (choose one)


chrisdembia:
Tested on Mac. LGTM.

image

aymanhab:
Thanks @chrisdembia

PR #1038, opened 2018-10-25
Remove vtk50 folder as now unused

aymanhab:
Fixes issue #0

### Brief summary of changes
Remove unused folder that used to contain vtk 

### Testing I've completed
None

### CHANGELOG.md (choose one)

- updated...


aymanhab:
Will use this cleanup no-op to trigger GUI build once change to opensim-core is done

PR #1039, opened 2018-10-25
Update for connectee path changes

chrisdembia:
This PR updates the GUI for recent changes in opensim-core (https://github.com/opensim-org/opensim-core/pull/2310). 

I also updated all the bindings; that will make this PR difficult to review. @aymanhab let me know if you want me to redo the PR so there are fewer changed files.

These changes resulted from various testing.

1. Loading a model, editing a component's sockets, saving the model, and ensuring the model file contained an updated connectee path. Tested this for the following:
  - Joint
  - Marker
  - BushingForce
2. Plotting with coordinates as the x value: ensuring the full path for the coordinate value state variable is used.
3. Running Scale, IK, ID, RRA, CMC for gait2354.
4. Editing the frame for fixed path points in the GeometryPath editor.

It is likely that more changes to the GUI are required. Anywhere where we use only a component's name to satisfy a connection will no longer work.

I wonder if some member variables in `SingleModelGuiElements` need to be updated to use absolute paths instead of simply names:
  - bodyNames
  - frameNames
  - physicalFrameNames
  - coordinateNames
  - unconstrainedCoordinateNames
  - actuatorClassNames
  - actuatorNames
  - markerNames


aymanhab:
@chrisdembia Thank you very much for getting this started. It may make sense to separate the Bindings upgrade and converting the calls into one commit or PR and then fixing the functionality in smaller ones so that we can distribute the testing. There're 663 files but a ton of them are SWIG generated using 3.0.12 vs. 3.0.9 Was that needed for this PR? If not I can recreate the files with 3.0.9 and will have a smaller set of changes to look into.
I didn't think that changes will be that big mainly because we don't use/find connections explicitly except in one single panel/editor (ConnectionEditor).  For now I'll try to get a feel for the nature of changes needed and get back to you. Thanks again 🥇

chrisdembia:
Thanks @aymanhab for improving the diff.

Gui/opensim/view/src/org/opensim/view/nodes/OneMarkerNode.java
```diff
@@ -199,6 +199,7 @@ public String getUndoPresentationName() {
        ExplorerTopComponent.addUndoableEdit(auEdit);
 
         obj.changeFramePreserveLocation(context.getCurrentStateRef(), PhysicalFrame.safeDownCast(comp));
+        obj.getSocket("parent_frame").finalizeConnection(obj.getRoot());
```
aymanhab: Can we move this inside the changeFramePreserveLocation method to avoid expicit finalizeConnection here? Also do we need the same for PathPoint changes?

chrisdembia: I know it's not clearly documented but I think `finalizeConnection()` should be called by API users. There isn't anywhere that the API calls `Component::finalizeConnections()` or `AbstractSocket::finalizeConnection()` internally (except `Model::initSystem()`).

As for PathPoints, I think I looked into this and it wasn't necessary.

aymanhab:
@chrisdembia I think the build is failing because master is not in sync with this branch. 
When all core PRs finish building, we should build locally against latest opensim-core, thenmerge this branch so that ci can make cross platform builds to assess.

chrisdembia:
Sure. But I think we can already try building locally against the latest opensim-core. I'll do that soon.

aymanhab:
@chrisdembia master build is failing due to trying to checkout this deleted branch, did we slip the change to use this branch to ci files by mistake?

chrisdembia:
I deleted the branch accidentally. Sorry.

chrisdembia:
Wait, why would master depend on this branch?

chrisdembia:
I think you can just rebuild the master branch. I just started one: https://ci.appveyor.com/project/opensim-org/opensim-gui

aymanhab:
> Wait, why would master depend on this branch?

Exactly why I'm 😕

PR #1040, opened 2018-10-26
[WIP] Support old states files

aymanhab:
Fixes issue #0

### Brief summary of changes
When loading states files from earlier Betas, column headers do not contain "jointset" for coordinates. 

### Testing I've completed
While I was able to associate the motion with the jumper model, motion doesn't play yet. Either because similar change needed for model.formStateStorage or storage.getStateIndex or on the GUI side/MotionDisplayer
### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/motions/MotionsDB.java
+++ b/Gui/opensim/view/src/org/opensim/view/motions/MotionsDB.java
@@ -194,8 +194,10 @@ public class MotionsDB extends Observable // Observed by other entities in motio
       int numCoordinates = cs.getSize();
       int numUsedColumns = 0;    // Keep track of how many columns correspond to Coords or Markers
       for(int i=0; i<numCoordinates; i++){
+          String testName = cs.get(i).getRelativePathName(modelForMotion);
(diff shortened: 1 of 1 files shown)


aymanhab:
@chrisdembia @jenhicks This is still work in progress but suggests work on the webinar may need to depend on data files from earlier betas.

chrisdembia:
I think the issue is in motionAssociationPossible. I vote for not supporting states files from earlier betas. If we agree on that, then merging the delay_connectee_name/path changes should resolve the issue.

aseth1:
> I think the issue is in motionAssociationPossible. I vote for not supporting states files from earlier betas. If we agree on that, then merging the delay_connectee_name/path changes should resolve the issue.

👍 
Also `MotionDB` should not be making these decisions. We should centralize to `Model`.

aymanhab:
@chrisdembia You edited the same line to handle the latest version,  I don't plan to merge this pending moving the checks down to the opensim-core side.

aseth1:
> @chrisdembia You edited the same line to handle the latest version, I don't plan to merge this pending moving the checks down to the opensim-core side.

closing ...

PR #1041, opened 2018-10-29
Expose visualizer port

aymanhab:
Fixes issue #0

### Brief summary of changes
This PR centralizes where port number for visualizer is set, and allows for overriding it from command line option in conf file.
### Testing I've completed
Built GUI and launched it and visualizaer worked as expected

### CHANGELOG.md (choose one)

- no need to update because new in 4.0


Gui/opensim/visualizationServer/src/org/eclipse/jetty/JettyMain.java
```diff
@@ -58,8 +69,9 @@ public static void main(String[] args) {
     public static void startServer() {
         try {
             if (!serverup){
+                getSERVER_PORT(); // set port based on command line if any
```
chrisdembia: From what I can tell, `getSERVER_PORT()` does not edit the `SERVER_PORT` static variable.

aymanhab: The static block in JettyMain is what's responsible for parsing the command line option and setting the static variable SERVER_PORT

PR #1043, opened 2018-11-01
Rather than name use full path as comment 

aymanhab:
This will allow output path to show at the bottom of Properties window, so it can be entered in OutputReporter by users

Fixes issue #0

### Testing I've completed
Loaded model, selected output and Fulll path displayed at the bottom

### CHANGELOG.md 

- We should update user guide to indicate how to use this OutputPath. Current format is not final pending opensim-core changes


aymanhab:
@chrisdembia This is the change I showed you so that you can test API changes to Outputs display in GUI

aymanhab:
Thanks @chrisdembia

Issue #1057, opened 2018-11-07
Commas appear in "Static Pose Weights" in Scale Tool


carmichaelong:
### Steps to reproduce
1. Open 2354
2. Load scale setup file
3. Go to Static Pose Weights --> R.ASIS (under markers)
4. Note that "Weight" has a comma in it: "1,000"
5. Try to change it to "8,000": doesn't work
6. Try to change it to "8000": works

### Expected result
Either the "Weight" box doesn't give the user a comma in the box OR commas can be accepted

aymanhab:
This is in the upper left corner of the panel and not in the table? Can you confirm? Thanks

chrisdembia:
Upper right corner of the panel, not in the table.

chrisdembia:
Verified on Mac.

PR #1068, opened 2018-11-08
Remove comma from format of large numbers

aymanhab:
Fixes issue #1057

### Brief summary of changes
Changed format string to not include commas

### CHANGELOG.md (choose one)


aymanhab:
@carmichaelong The format strings (with commas) were used across the board in tools and plotting etc. You can test the artifact to see if the problem was fixed when you have a chance.

carmichaelong:
LGTM. Checked that Scale and IK both have the commas removed, and thus it's easier to change the numbers. You mentioned plotting, but saw some commas along the y-axis:
image

This is definitely not an issue in this PR, and I don't think it's actually an issue in the larger context either, so I'd say it's up to the team whether or not to open another issue.

chrisdembia:
I'm fine with the commas in the plotter.

PR #1069, opened 2018-11-08
Undo refactoring unit conversion of marker data

aymanhab:
Fixes issue #0

Can't associate marker data trc file with IK motion in tutorial 3

### Brief summary of changes
Undo seemingly innocent refactoring to unit conversion, that resulted in marker data not showing altogether

### Testing I've completed
Ran IK, loaded motion, associate marker data, and force data successfully and animated
Loaded trc files in (meters) and in mm and played them back without an issue.

### CHANGELOG.md (choose one)

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/experimentaldata/AnnotatedMotion.java
+++ b/Gui/opensim/view/src/org/opensim/view/experimentaldata/AnnotatedMotion.java
@@ -120,7 +120,7 @@ public class AnnotatedMotion extends Storage {
                 getBoundingBox()[4]+", "+
                 getBoundingBox()[5]
                 );
-        for (int i=0; i<6; i++) getBoundingBox()[i] *= getUnitConversion();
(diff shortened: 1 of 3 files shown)


aymanhab:
Thanks @chrisdembia as discussed the issue was not that users can't associate but that the Markers don't show up.

chrisdembia:
Is this ready to test?

aymanhab:
Yes, ready to test as soon as PR build finishes.

chrisdembia:
image


This PR fixes the issue but I am not sure how to test for other possible side effects. Any suggestions?

aymanhab:
Thanks @chrisdembia for the quick turnaround.
Looking at the code diffs you'll see it only affects unit conversion (it was suggested that using multiplication rather than division is a straightforward change that has no side effects 😃 )
The functionality that I would check is being able to load MarkerData (to preview, and to associate) and to change the sizes of Experimental Markers, and same for ForceData. If all works then we're good.  Thank you 👍

PR #1071, opened 2018-11-08
removed duplicate scripts

jimmyDunne:
Fixes issue opensim-org/opensim-models#102

### Brief summary of changes
The Scripts that are used for the metabolics example have been moved to opensim-models and should be deleted here to avoid duplication

### CHANGELOG.md (choose one)
- no need to update because it is a file rearrangement


aymanhab:
Not sure if the branch will be reused vs. deleted and if other changes to models repo are coming, so I'll do neither, just let me know @jimmyDunne @chrisdembia

chrisdembia:
I'm pretty sure the branch is no longer needed. If it is, @jimmyDunne can push back his local copy.

Issue #1052, opened 2018-11-07
Plotting can change position of the model


jenhicks:
### Steps to reproduce
Load the wrist model
Plot the flexion moment arm as a function of flexion for the pre/post ECU
Close the plot and change the elbow flexion angle.

### Expected result
Wrist flexion does not change

### Actual result
Wrist flexes

(include screen grabs where appropriate)

### Environment and GUI version
Mac OpenSim 4.0-2018-11-06-f854814

carmichaelong:
Replicated issue with arm26 and then tested that this was fixed in OpenSim 4.0-2018-11-08-ca20f56c.

@chrisdembia you mentioned this before, should there be another issue made for this (for after 4.0)?

> Another related issue: if I have loaded a motion with my model and then I put the model in some random pose and then open any of the tools or the plotter window, the model goes to the initial pose in the motion.

chrisdembia:
Done https://github.com/opensim-org/opensim-gui/issues/1077

PR #1072, opened 2018-11-08
Save/Restore state around running the analyzeTool (by the plotter)

aymanhab:
This is necessary since the GUI and tool share the same State reference.

Fixes issue #1052

### Brief summary of changes
Save/restore the State's Y vector so it doesn't get polluted by the plot/analyzeTool run.

### Testing I've completed
Tested workflow in bug report and got correct behavior

### CHANGELOG.md (choose one)


aymanhab:
All tools (other than IK) create a copy of the model so this is not an issue. Only plotter doesn't clone/copy for historical/efficiency purposes. Should be easy to fix post 4.0, or even better, move to using Outputs instead of analyzeTool,

aymanhab:
Thanks @chrisdembia

Issue #1061, opened 2018-11-07
SO: tool fails the first time in the Static Optimization Example


chrisdembia:
SO example: Static Optimization tool crashes the first time it is run (followed steps, saved settings file, then ran -> "tool execution failed or canceled by user"). Then when you close static optimization and reopen it, and load the saved settings file, the tool works. 

https://docs.google.com/spreadsheets/d/1E9Em79S5Y52hfAePmK-XcvTcPYwFWQBSchYemuWYTfE/edit#gid=1519816561

nickbianco:
Confluence example: https://simtk-confluence.stanford.edu/display/OpenSim/Working+with+Static+Optimization

nickbianco:
@carmichaelong will narrow down the issue in the SO confluence example

aymanhab:
There're two issues at play here:
1. Loading the data file for external loads, this has been moved to opensim-core and boils down to trying to load the file while in the wrong folder.
2. GUI not accounting for SO failing and it tries to load activations to color muscles regardless!

Issue 2 is fixed by PR https://github.com/opensim-org/opensim-gui/pull/1075

chrisdembia:
I am starting to look into the opensim-core fix. Feel free to take it over for me though; I'm not sure hwo far I'll get.

aseth1:
In 1. the data is loaded the second time around by the API, so why does it fail the first time?

It seems like there might be a difference between the working directory when editing the Setup in the GUI (failure) versus expecting the Tool (ExternalLoads) to load the data (which works). My sense is that when editing, the GUI either checks for existence and/or changes the working directory. When running the Tool with a complete setup (no editing) that check/change is not occurring.

chrisdembia:
I think that the pathway, below, of changing the working directory to the location of the ExternalLoads XML file is never exercised, because we are always operating on a copy of ExternalLoads (in the model) that no longer has a document.

https://github.com/opensim-org/opensim-core/blob/2c60796f4ebd4ad50cdee23d0ff99cd73ec67d2a/OpenSim/Simulation/Model/ExternalLoads.cpp#L184-L199

aseth1:
Yes we clone the ExternalLoads because we cannot have the same ExternalLoads live be part of two models. We need a secondary model to perform the transformation using kinematics for external loads.

[...]

PR #1075, opened 2018-11-08
Before trying to load SO results (activation) 

aymanhab:
check that there are valid results before trying to visualize them

Fixes issue #1061 partly

### Brief summary of changes
Before loading activations, make sure the storage containing them is not null

### Testing I've completed
Run through issue steps and now only get first error due to missing/failed-to-locate force-data file.

### CHANGELOG.md (choose one)


chrisdembia:
Verified locally that I do not get a red circle error after not finding the ExternalLoads.

Issue #1060, opened 2018-11-07
Using GUI to change marker positions: new positions are not saved


chrisdembia:
https://docs.google.com/spreadsheets/d/1E9Em79S5Y52hfAePmK-XcvTcPYwFWQBSchYemuWYTfE/edit#gid=1519816561

aymanhab:
@chrisdembia I just loaded arm26, saved it in 4.0, changed the location of a marker and saved and the changes were saved. There must be a specific scenario to reproduce. Any details? @carmichaelong can you reproduce? Thank you

chrisdembia:
@carmichaelong let me know if you want to work together on reproducing this.

aymanhab:
For the record, I tested only editing the numbers in the Properties panel and not other ways (e.g. Graphically dragging or changing the Frame the marker is attached to, or loading Markers from file and editing or creating a new Marker and edit 😸 ) Would be good if a witness can help us narrow the search path.

chrisdembia:
The issue was specifically moving the marker graphically. Sorry that wasn't clera.

chrisdembia:
I am having some difficulty reproducing the situation.

Rachel was also saying the issue is sporadic; it doesn't always happen.

Whether or not a marker is fixed can also complicate the matter.

@rwj1 could I come by the HPL to observe the bug?

carmichaelong:
@chrisdembia FYI I also haven't been able to reproduce yet.

aymanhab:
@chrisdembia @carmichaelong @jenhicks I'm woking on a fix to refresh the text faster and allow undo, but it is not clear if related or if it should be in this last sprint. Let me know your thoughts.

aymanhab:
Nice investigative work @chrisdembia 
Loading markers from a file has changed a few times including last time when we moved MarkerSet to be component, every Marker addition likely goes through the expensive initialization sequence (though 20 seconds is too much and I'm sure we can optimize). 
In step 5, do they get different behavior if they click in the Navigator or elsewhere in the application?
Is it the case that one of the 3 markers didn't move while the two others did? Just clarifying

chrisdembia:
Lauren sent me her own files and I still could not reproduce the issue. The markers did take a long time to load on my computer.

Could the issue be related to the speed of the computer? Or Windows vs Mac?

aymanhab:
[...]

PR #1080, opened 2018-11-10
Marker location update

aymanhab:
Fixes issue #1060 
### Brief summary of changes
Instead of setting the location property of Marker thru graphical edit directly, invoke the PropertyEditor workflow to updated location so users get feedback immediately.

### Testing I've completed
Editing markers graphically updates properties window immediately now, likely avoid the situation leading to 1060 where graphical display and numbers are not in sync.. 

### CHANGELOG.md (choose one)


carmichaelong:
I used OpenSim 4.0-2018-11-08-3836df48 as a comparison point. In that version, the location update wasn't slow, but there was definitely a pause of ~0.5 seconds. With this branch, the update in location seems almost instantaneous.

chrisdembia:
I tested on Mac. The update seemed fast though I did not compare to a previous build. Undo/redo seemed to work for the most part.

PR #1089, opened 2018-11-13
AppVeyor: upload artifacts to Google Coud

chrisdembia:
With this PR, artifacts are uploaded to Google Cloud instead of Azure, to improve download time.

We also upload a ZIP of the GUI in addition to the installer. If you use a program like 7-Zip, you can unzip the ZIP much faster than the installer, which could be good for testing.


chrisdembia:
Can we merge this?

aymanhab:
If failures are just cancelled builds and the deploy thru google cloud works then please feel free to merge.

chrisdembia:
That is the case.

Issue #939, opened 2018-08-03
Ubuntu build script


idhamari:
For a clean Ubuntu 18.04, writing and testing bash script to:

1. Download and Install all required libraries e.g. gcc, java, netbeans, etc.
2. Download opensim-core and opensim-gui 
3. Build opensim-dependencies 
4. Build and install opensim-core
5. Copy required files for opensim-gui 
6. Build opensim-gui

So far, I completed the first 4 parts (still need more testing). It is not clear how to process step 5, what are the sources and destination files for these projects CopyOpenSimCore, CopyVisualizer, CopyModels, CopyJRE,PrepareInstaller? it would be nice if you put the paths here.

chrisdembia:
Refer to the `.travis.yml` script.

idhamari:
If I am java expert I would not ask this question. It is much easier if you put in a simple copy commands. I am not sure why the need for a visual studio or cmake project to do that when simple batch or shell script can do it.  

I found a work around by using the files after the build, the compressed size is around 500 MB. so it is still downloadable.  Testing in progress ...!

chrisdembia:
FYI We are less likely to use a script that does not conform to our current practices, unless there are clear advantages.

idhamari:
Sure! 

This [script](https://github.com/opensim-org/opensim-gui/files/2259383/buildOpenSim.zip) is tested in a new Ubuntu 18.04. I zipped the source-codes and upload it to our server (if there is a branch for a stable source-code, it can be used instead). Feel free to modify the script or re-upload the zipped files based on your need. If there is any legal conflict, please let me know and I will just remove the shared links. 

Next step: if there is no binary packaged already for Ubuntu, I will try to figure out how to make one.

chrisdembia:
Thanks @idhamari! I've downloaded your scripts. We will try to take a look when we get a chance.

chrisdembia:
Keep in mind that we still have no timeline for releasing on Ubuntu.

AlbertoCasasOrtiz:
@aymanhab Yes, this is the final place for build instructions. This way we can follow all of the changes in the wiki.

PR #1098, opened 2018-11-21
Commit build_on_ubuntu.sh from @idhamari

chrisdembia:
In #939, @idhamari shared a build script for the GUI. I do not think we are ready to support Ubuntu but I thought it would be good to store his script in the repository for future reference. @idhamari is that okay with you?


idhamari:
Sure it belongs to everyone!

aymanhab:
Considering PR #1181 I think we can close this issue, or do we still this need merging this PR @chrisdembia ?

chrisdembia:
I don't want us to lose this file, so I think we should still put it somewhere on the master branch, with whatever disclaimers are appropriate.

aymanhab:
There's a new script now on the wiki by @AlbertoCasasOrtiz so this issue is completed, please open otherwise

PR #1099, opened 2018-11-21
Bump version to 4.1.

chrisdembia:
See https://github.com/opensim-org/opensim-core/issues/2170

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Allow specifying VERSION for the project.
 cmake_policy(SET CMP0048 NEW)
 
-project(OpenSimApp VERSION 4.0)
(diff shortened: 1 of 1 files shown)


chrisdembia:
@aymanhab this version is how we set app.version; https://github.com/opensim-org/opensim-gui/blob/87398e2c07820fe10e26713196f03e88ceb48f1d/Gui/opensim/CMakeLists.txt#L139

So for the purpose of building a distribution, this is intended to be the single place where the version number is set and used.

aymanhab:
Thanks for the clarification @chrisdembia though the block you pointed to is only for mac installer. Would be good to unify the mac and windows at some point. unrelated to this PR.

PR #1110, opened 2019-04-01
Allow non Muscle components that contain path to provide colors

aymanhab:
Fixes issue #0

### Brief summary of changes
GUI used to treat Muscles as special owners of GeometryPath to get colors during simulation/animation. This is unnecessray and prevents custom components (e.g. Ligaments) from providing state dependent colors.
### Testing I've completed
Loaded model with a ligament and allowed it to run FD, display turned gray rather than state-independent green. Will be useful for Smith2016Ligament class.

### CHANGELOG.md (choose one)

- updated...


aymanhab:
Folded into PR #1142

Issue #1105, opened 2019-02-22
GUI is adding offset to experimental data  (it shouldn't be)


jimmyDunne:
### Steps to reproduce
(1) Preview experimental marker data in GUI. 
(2) Preview corresponding force data in GUi. 
(3) Sync motions. 

### Expected result
Marker and force data are displayed in the exact, global, location that they are defined in the experimental data files.

![image](https://user-images.githubusercontent.com/5296507/53247177-b57b9600-3680-11e9-8523-835ca9d27294.png)

### Actual result
The GUI adds an offset to the view frame of the force data and so it appears as if there is a real offset between the marker and force data. 

![image](https://user-images.githubusercontent.com/5296507/53247153-a399f300-3680-11e9-88e7-d96c83c15414.png)

I would guess that this behavior occurs because the the Experimental Data Previewer is sharing some of the same functions as the Model Opener, but this particular behavior is incorrect. Experimental data should never have offsets automatically injected into the visualization since this makes it difficult for the user to know what is exactly wrong. Case in point is this [forum post](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=10075&p=0&start=0&view=&sid=f5699923b82824de2e6dbe358e0e8ae6). 

### Environment and GUI version
Mac and Windows

jenhicks:
This has come up a couple times on the forum (two users).

jimmyDunne:
Fixed in #1118

PR #1118, opened 2019-06-13
Pass modelname2visualizer

aymanhab:
Fixes issue #1105 

### Brief summary of changes
Visualizer (now responsible for adding visuals/offsets) now respects ExperimentalData (models). Test was there already but didn't work because of late scene graph changes in 4.0.

### Testing I've completed
Loaded trc, grf data for Gait10dof18musc model, synchronized them and all showed with no offset. Also tested that we're able to move using display_offset.

### CHANGELOG.md (choose one)
- Updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,4 +7,5 @@ request related to the change, then we may provide the commit.
 v4.1
 ====
 - Fix issue #1112: TRC file written by GUI cannot be read by TRCFileAdapter 
-- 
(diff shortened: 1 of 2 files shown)


aymanhab:
@jimmyDunne Artifact ready for testing

jimmyDunne:
I downloaded and installed on both Windows and Mac. The reported issue, experimental data having an offset applied, seems to be fixed 🎉 👍

aymanhab:
Thanks @jimmyDunne, much appreciated

Issue #1115, opened 2019-06-08
Geometry Path has default appended upon launch despite what user had set


aseth1:
Likely related to #745 

### Steps to reproduce
1. launch OpenSim4.0 GUI and  set `Paths: Geometry Search Path` to something, say <My_Prefered_Path>
2. Open a model that uses geometry in said path. All good.
3. Quit and relaunch OpenSim4.0 GUI and reload your model
4. Model looks janky because it is reading default geometry that it finds in the default path for Geometry 

### Expected result
![image](https://user-images.githubusercontent.com/6494122/59139787-6a914000-894a-11e9-8b2d-cc6dff0aeda3.png)

### Actual result
Default OpenSim Geometry is magically appended and takes precedence over the path I set last time GUI was running and all was well and now my model looks awful.

![image](https://user-images.githubusercontent.com/6494122/59139837-f7d49480-894a-11e9-903b-b5afdf1aa103.png)

### Environment and GUI version
OpenSim 4.0 Win64

aymanhab:
Good find @aseth1 👍

aseth1:
closed by #1119?

aymanhab:
Thanks @aseth1 and @jimmyDunne

PR #1119, opened 2019-06-19
No add default geometry search path on launch

aymanhab:
Fixes issue #1115 

### Brief summary of changes
On launch, geometry search path is not modified unless blank.

### Testing I've completed
Modified geometry search path, exit, relaunch, search path stay same.

### CHANGELOG.md (choose one)


aymanhab:
I'd just test the behavior in the artifact @jimmyDunne . Not sure this was ever an issue on Mac

jimmyDunne:
Thanks, @aymanhab. I downloaded and installed the artifact, testing on W10 VM. 

It seems like the Geometry path is persistent between launches.

If the path is empty, on re-launch the path will be updated (as expected)

Everything looks good!

aymanhab:
Thanks @jimmyDunne 👍

Issue #1123, opened 2019-07-05
Disabled muscles show in red when playing back SO results


aymanhab:
### Steps to reproduce
Load Arm26 and static optimization default setup, run analysis, disabled muscles show in red

### Expected result
disabled muscles show a different color outside the blue--to-red spectrum
### Actual result

(include screen grabs where appropriate)

### Environment and GUI version
4.0

aymanhab:
Some confounding factors worth discussion before a solution is implemented:
- GUI has 2 modes: mode 1 does not color based on state at all as in load/edit, mode 2 muscle color is based on state as in motion playback.
- SO uses a mechanism that overrides the color from state to color by activation from a file/storage (due to the fact that SO produces no consistent state to use).

tkuchida:
@aymanhab Perhaps "user stories" would be helpful here. There are countably many scenarios in which muscles are drawn, and presumably one could determine what the user would want to see in each case. IIRC there are several GUI "modes" (model building vs. simulating vs. playback of results). Perhaps these modes should be communicated to the user (e.g., by changing the color/look of the interface)?

aymanhab:
@tkuchida Good point. I thought about that during 4.0 but decided against it because there were only 2 modes (simulation and playback are intended to produce the same visuals). When we introduce model building it could help make it clear when a model is under-construction vs. ready vs. simulating at that point, but I'd like to hear your rationale for doing it now as well.

tkuchida:
> I'd like to hear your rationale for doing it now as well

![](https://user-images.githubusercontent.com/4203505/60750677-442ae880-9f7a-11e9-85c0-aafe71be27a4.png)

Perhaps the GUI should be communicating some sort of "metadata" (mode/condition/state) to the user. For example, how would a user visually distinguish between an IK solution (all muscles always red) vs. a dynamic simulation with maximally activated muscles (all muscles always red)?

aymanhab:
@tkuchida I agree we should communicate more info to the user, the trick is to do it without overwhelming the user with unnecessary or repeated details. ColorBy can be displayed in the status bar or messages window when it changes, but some are implicit as when you run FD or playback. IK is a bit more difficult since we use formStateStorage which treats all motions the same regardless of contents.

aymanhab:
I made a change so that if we color by activation from file (as in SO) then muscles that are not found (e.g. disabled) are colored Yellow (a color outside the spectrum of blue-red). Some hanging issues to discuss though:

[...]

PR #1125, opened 2019-07-11
Color by activation handle disabled muscles

aymanhab:
Fixes issue #1123
### Brief summary of changes
Modify color maps to color muscles with a special color outside blue-to-red spectrum to indicate disabled.
### Testing I've completed
Ran SO with disabled muscle, got correct behavior/coloring.

### CHANGELOG.md (choose one)

- updated...


aymanhab:
@jimmyDunne you can test this. First run SO with some disabled muscles, then use "Color Muscles from File..." and pick activation from SO.

jimmyDunne:
Thanks, Ayman. I download the latest artifact and installed on a Windows 10 VM. I ran Static Optimization with, and without, the soleus_r muscle enabled. 

During Static Optimization execution, when soleus_r _enabled_ was unchecked, the muscle was colored grey. Immediately at the end of Static Optimization, the muscles were colored blue, with the soleus being red. 

Screen Shot 2019-07-17 at 2 02 42 PM

When I moved the motion slider, the muscles switched to the colored states from Static Optimization, including the greyed out soleus

Screen Shot 2019-07-17 at 2 03 23 PM

All other times, the coloring was working great and seemed to be what we would expect. The Static Optimization output files did not include the soleus_r muscle and, when coloring muscles using an activation file, the soleus was greyed out. Seems to work great except for that one little glitch described above. 

Let me know if there is anything else I can check?

aymanhab:
Thanks @jimmyDunne I think this is due to the GUI being caught in between rendering live results and coloring with results from file which may have been always there but became more obvious once we switched to colors outside the blue-red spectrum. Will let you know.

aymanhab:
@jimmyDunne, what model, files did you use. I get similar glitches but that's because the time window for the SO result files are different from those of the motion. Thank you.

aymanhab:
@jimmyDunne, please retry the build after this commit. Thanks for testing

jimmyDunne:
@aymanhab. I downloaded, installed, and tested with the latest artifact. I above issue (incorrect color when SO has finished) seems to be fixed 🚀 

 I tested loading states again and it still is working great. LGTM 🥂

aymanhab:
Awesome, thanks for testing.

Issue #1111, opened 2019-05-02
Muscle wrapping visualization is incorrect when muscle loses contact with WrapCylinder


tkuchida:
### Steps to reproduce
- open Rajagopal2015 model
- change knee_angle_r to 30 degrees

### Expected result
Gastroc path looks reasonable.

### Actual result
![](https://user-images.githubusercontent.com/4203505/57107787-f5a16980-6cfe-11e9-984e-f3ce0921e1c2.png)

Path seems to be computed correctly:
![](https://user-images.githubusercontent.com/4203505/57107802-0651df80-6cff-11e9-96d2-fd8f2831ee3b.png)

### Environment and GUI version
Public release of 4.0

jenhicks:
@aymanhab will investigate, given recent discussion on forum.

aymanhab:
I'm working on a fix now,

aymanhab:
Let me clarify the scope of the issue:
- This has no effect whatsoever on computation or correctness of results, pure visualization issue.
- All models display correctly on loading regardless of wrapping. If wrong display happens save the model in this bad configuration and load it and it will be fine.
- Only models/muscles that could be affected are those with multiple wrap objects when an object switches from engaged to disengaged while some other objects are engaged either during a motion playback or while moving a coordinate slider.

jenhicks:
@jimmyDunne Can you test this one?

jimmyDunne:
Placing the model in the affected pose, saving, and reloading;

![image](https://user-images.githubusercontent.com/5296507/66132578-46c1cc00-e5c3-11e9-84f0-0651e4ace733.png)

aymanhab:
The behavior you described is exactly how it was before the fix, that's not
what I see using the branch with the PR that has not been merged into
master yet. Can you report what environment did you use? Thank you

jimmyDunne:
I am on Mac. I used the OpenSim-fix_visualization_multiple_wraps artifact

PR #1134, opened 2019-09-13
Fix place where  pathpoints are shown when not engaged.

aymanhab:
Fixes issue #1111

### Brief summary of changes
Path visualization now keeps track of all wrap-pathpoints that are engaged so it can decide what to do when a wrap-object becomes disengaged.

### Testing I've completed
Loaded model reported on forum and in issue 1111 and moved sliders, loaded motions, all displayed correctly.

### CHANGELOG.md (choose one)

- no need to update because rare scenario to reproduce, issue fix could be included


aymanhab:
@jenhicks @jimmyDunne suggested fix works for Apoorva's model but breaks two muscles in our Shoulder model, investigating a more robust fix before passing to others to test.

aymanhab:
@jimmyDunne I believe I have this issue fixed along with #2569 would be good to give it a good spin, especially for muscles with models with multiple wrapping (Apoorva's , Arm26 and Shoulder model) as well as the files in issue reports once a GUI build is available. Thank you

aymanhab:
James, to test if you have the right build load Apoorva's model and load the motion attached here
[bad_gas.txt](https://github.com/opensim-org/opensim-gui/files/3688178/bad_gas.txt) if gaslat and gasmed show correctly in first from then you have the correct build otherwise you don't.

jimmyDunne:
Seems to show the expected behavior on opening 🍾 
image

Issue #1107, opened 2019-02-28
Marker shading is inconsistent with the shading of all other geometry in the 3D view


aseth1:
### Steps to reproduce
Load any model with markers and look at it.

### Expected result
Spheres are shaded with light source in common with all other objects.

### Actual result
Spheres appear to be magically illuminated from below (that is from the floor up).

![image](https://user-images.githubusercontent.com/6494122/53605727-4a90fa00-3b6d-11e9-824a-a9d51459dad6.png)

### Environment and GUI version
OpenSim 4.0 on Mac and Windows

tkuchida:
There was an old issue about this. IIRC, the issue was that the back faces were being drawn rather than the front faces. It looks like an open hemisphere is being drawn, and the light from above the model is reflecting off the inside of the hollow bottom.

PR #1140, opened 2019-09-30
Make model marker less shiny

aymanhab:
Fixes issue #1107
### Brief summary of changes
Change material specification of model markers to match those of experimental markers for consistency. Along with opensim-core fix to scale factors these two fixes combined result in better looking model markers.
### Testing I've completed

Created new markers, loaded models with markers and inspected visualization

### CHANGELOG.md (choose one)
-updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,4 +10,5 @@ v4.1
 - Fix issue [#1105](https://github.com/opensim-org/opensim-gui/issues/1105): GUI adding offset to Experimental Data by default
 - Fix issue [#1123](https://github.com/opensim-org/opensim-gui/issues/1123): GUI displays disabled muscles in red-blue range during/after StaticOptimization.
 - Fix issue [#1127](https://github.com/opensim-org/opensim-gui/issues/1127): Change default time in the toolbar forward simulation tool to 5 seconds. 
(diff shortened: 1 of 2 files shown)


aymanhab:
@jimmyDunne If you review and merge will get an Artifact of the gui with integrated fix.

jimmyDunne:
They are definitely different. I have a few comments just as impressions as I looked at it for the first time. 

From a not-so-far distance away, the markers appear more 'black' than pink. 
image

If I would have to guess, it may because the dark shading on the marker is on the superior surface of the maker, rather than than the infererior. The experimental markers seem to have the darker shading on the inferior surface of the sphere, which we may want to mimick. I think that would help the Model marker pinl coloring to stand out. 

image

aymanhab:
James, the wrong shading suggest this build of the gui does not have the
core fix. I can restart the gui build to force that unless you locally
built latest master on both core and gui.

jimmyDunne:
I don't build the GUI locally so if you could force another CI build I can get the artifact?

aymanhab:
Thanks @jimmyDunne Artifact of master is ready to test. Sorry about the confusion

jimmyDunne:
Yep, looks great! Thanks, Ayman. 
image

Issue #1139, opened 2019-09-27
Sphere color doesn't change when path actuator color is changed


tkuchida:
### Steps to reproduce
- open a model containing a path actuator (e.g., [gait2354_simbody_withPathActuator.osim.txt](https://github.com/opensim-org/opensim-gui/files/3664791/gait2354_simbody_withPathActuator.osim.txt))
- in Navigator, right-click on "my_path_actuator" -> Display -> Color…
- change color

### Expected result
Color of all associated geometry changes.

### Actual result
Cylinders change color but spheres do not.

![image](https://user-images.githubusercontent.com/4203505/65807035-a569ee80-e15a-11e9-9fe7-db4ba58bf640.png)

### Environment and GUI version
Public release of 4.0.

PR #1142, opened 2019-09-30
Fix path actuator visualization 1139

aymanhab:
Fixes issue #1139
### Brief summary of changes
GUI code that handles path appearance changes was strictly checking for Muscle subclasses while capable of supporting any Force that hasGeometryPath() 

### Testing I've completed
Loaded model in issue 1139 and changed color/appearance/show/hide/undo/redo of Appearance changes and all performed as expected.
Would be good to have a model with other subclasses (e.g. Ligament, CluchedPathSpring) to test behavior.

### CHANGELOG.md (choose one)

- Will update before merge if we agree it belongs to 4.1


aymanhab:
undo color change doesn't work, all other operations work on both PathActuators and Ligaments. Looking into it.

aymanhab:
Issue with undo found and fixed, now ready for review.

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -790,15 +791,15 @@ public void appendToFrame(JSONArray bodyTransforms_json, boolean colorByState, J
                 UUID pathUUID = pathList.get(geomPathObject);
                 JSONObject pathUpdate_json = new JSONObject();
                 pathUpdate_json.put("uuid", pathUUID.toString());
-                if (Muscle.safeDownCast(geomPathObject.getOwner())!= null){
+                // Allow non-muscles to provide colors if (Muscle.safeDownCast(geomPathObject.getOwner())!= null){
```
chrisdembia: Remove?

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -920,7 +921,8 @@ public JSONObject createAppearanceMessage(Component mc, AbstractProperty prop) {
             JSONObject commandJson = CommandComposerThreejs.createAppearanceChangeJson(prop, objectUuid);
             guiJson.put("command", commandJson);
             String commandName = (String) commandJson.get("name");
-            if (Muscle.safeDownCast(mc)!=null && commandName.equalsIgnoreCase("SetVisible"))
+            Force mcAsForce = Force.safeDownCast(mc);
+            if (mcAsForce!=null && mcAsForce.hasGeometryPath() && commandName.equalsIgnoreCase("SetVisible"))
                commandJson.put("type", "SetValueCommandMuscle");
```
chrisdembia: It seems like the visualizer knows only of muscles (based on the string "SetValueCommandMuscle"). It seems that the visualizer should handle paths, not muscles. Is this simply a matter of naming?

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -1344,7 +1346,8 @@ public void propagateGeometryPathCommandsToPathPoints(Muscle muscle, AbstractPro
         }
         else {
             JSONObject pathpointCommand = (JSONObject) lastCommand.clone();
-            GeometryPath gPath = muscle.getGeometryPath();
+            
+            GeometryPath gPath = GeometryPath.safeDownCast(force.getPropertyByName("GeometryPath").getValueAsObject());
```
chrisdembia: I would prefer avoiding `getPropertyByName()`. How do you know for sure here that the property exists? Should there be an if-statement?

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
aymanhab: Sure.

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
aymanhab: Exactly it's naming, muscles mean nothing more than a path to the visualizer.

chrisdembia: Can you create an issue for renaming/refactoring to improve the clarity of the code?

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
aymanhab: This is tricky because the presence of a GeometryPath in a Component is not built into the class hierarchy (so we can downcast and get/updGeometryPath). There's a method to check if a Force has one (hasGeometryPath) but no generic way to get it if the result is true. As it stands, the GUI calls this code only on Forces that have been checked to have GeometryPath by construction but I'm open to suggestions.

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
aymanhab: The JS object on the visualizer side though is called SkinnedMuscle so the language on the visualizer side is consistent, but feel free to open an issue regarding clarity of this or any other piece of code you review on the gui side. Thank you @chrisdembia

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
chrisdembia: Thanks for explaining. There is no guarantee that the force contains a property named GeometryPath. So there should be an error check.

A more maintainable solution would be to add a virtual `Force::getGeometryPath()`.

chrisdembia: For now, if you add an exception check and create an issue for `Force::getGeometryPath()`  I'm willing to approve the issue.

aymanhab: Test and exception added.Thanks @chrisdembia

aymanhab:
@tkuchida or @jimmyDunne would you be able to test the Artifact when available? Thank you

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
aymanhab: issue https://github.com/opensim-org/opensim-core/issues/2587 has been opened to address the getPropertyByName concern,

jimmyDunne:
Tested artifact— path points change color with line. 

image

aymanhab:
Awesome, thanks @jimmyDunne I've been testing with Ligament, if you can also test with ClutchPathSpring that would be great.

jimmyDunne:
Coloring seems to work well for ClutchPathSpring's

aymanhab:
Feedback addressed, now ready for review @chrisdembia

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
aymanhab: pinging @chrisdembia for review

aymanhab:
@chrisdembia would you please let me know if your feedback has been addressed. Thank you

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -1346,11 +1344,15 @@ public void propagateGeometryPathCommandsToPathPoints(Force force, AbstractPrope
         }
         else {
             JSONObject pathpointCommand = (JSONObject) lastCommand.clone();
-            
-            GeometryPath gPath = GeometryPath.safeDownCast(force.getPropertyByName("GeometryPath").getValueAsObject());
-            
-            pathpointCommand.put("objectUuid", getFirstPathPointUUID4GeometryPath(gPath).toString());
-            commands.add(pathpointCommand);
+            if (force.hasGeometryPath()){
+                GeometryPath gPath = GeometryPath.safeDownCast(force.getPropertyByName("GeometryPath").getValueAsObject());
```
chrisdembia: It is still possible for there to be an exception here. There is no guarantee that `hasGeometryPath()` returning `true` means there is a property named `"GeometryPath"`.

aymanhab:
Thanks for the review @chrisdembia 👍 as I commented above hasGeometryPath and getGeometryPath use the same mechanism so normally one implies the other but this can be addressed along with https://github.com/opensim-org/opensim-core/issues/2587. Will merge by the end of the day unless I hear otherwise from you.

PR #1145, opened 2019-10-28
IMU visualization

aymanhab:
Fixes issue #0

### Brief summary of changes
Load IMU quaternions into the GUI as "Sensor Data", and show Frames corresponding to IMUs in IMU world, allow users to specify configuration for visualization, or apply a Transform

### Testing I've completed
Loaded data files controlled visualization, offsets

### CHANGELOG.md (choose one)

- Will update once functionality ironed out


aymanhab:
@carmichaelong @malborno When you have a chance you can test the functionality and/or pass it on to other interested parties. Still need to tidy up comments and undo unrelated changes before review.

aymanhab:
@antoinefalisse If you can give this a spin and provide some feedback when you have a chance that would be great.

carmichaelong:
@aymanhab Tested with some of the testOpenSense files. Looks like it's working as intended. I could choose the different setups, and also loaded it with a current model as well. I was also able to use "Show Only" on individual sensors to look at that separately which is nice. I tested with the RHJC trial, and saw much more motion on IMUs on the right side of the body compared with the left.

aymanhab:
Hi Antoine,

NetBeans changed drastically after version 8 (developed while under Oracle)
as it became an OpenSource project maintained by Apache. You can try using
the latest version but it could be a major undertaking, if you do it would
be helpful though. The ci builds the GUI for you using NetBeans 8.2 so you
don't need to do it yourself, i mainly wanted to get feedback on the
functionality. Your call.

antoinefalisse:
Oh I see, thanks. So how can I download the GUI with this PR to test out the functionality? Sorry, not very familiar with CI.

aymanhab:
You typically go to the top level github page
https://github.com/opensim-org/opensim-gui
Then select "Actions" from the tool bar ("Code" "Issues"...."Actions"...)
This will show you a list of ci builds, scroll down to "IMU visualization"
https://github.com/opensim-org/opensim-gui/actions/runs/169015148
Near the top of the page there will be artifacts that you can download and
run as our normal installers.

Please let me know if you have any questions.

antoinefalisse:
@aymanhab I was able to visualize my own sensor data as well. Two visualization options did not work though:

- Place at matching segments on current model -> I got the following warning: "The following sensors (names) could not be matched to model, will be displayed at Origin".
- Attach to matching segment on current model -> I got an exception: "Not supported yet".

The IMU names in the orientation file match the `PhysicalOffsetFrame` names in my osim file. Not sure whether those options should work.

aymanhab:
Thanks for testing @antoinefalisse, the option to "place at matching segments" looks for "Segments/Bodies" in the current Model of the application and attaches the Frames at the corresponding Center of Mass. Obviously this was not documented, sorry but please let me know if it works now (with this in mind). We can allow for OffsetFrames as well instead of Bodies but the conjecture was that users would exercise this before they calibrate the model and place IMUs on it, however we can try to support both just let me know based on your use case.

The last option is not supported yet (Not sure if it will be needed as opposed to an "associate sensor data"). Thoughts?

antoinefalisse:
Hey @aymanhab I figured that you first need to close the sensor data you were previewing if you want to use the option "place at matching segments". That's why I had that warning last time. For the other options, you don't need to do that. It might be annoying for users to have to close what they already loaded in the navigator. 

I'm not sure what was initially intended with the currently not supported option. Was the idea to be able to load a motion and to have the previewed sensor data moving with the skeleton? If so I guess that would be quite nice.

aymanhab:
Thanks @antoinefalisse We can allow users to choose from currently open models in the application but this complicates the workflow so I lean to not do it, and instead issue a warning if the current model in the application is an ExperimentalData only model and the option was selected, or maybe disable in this case.

Yes I'm leaning to remove the last option from here and instead add the option to associate sensor data which would be good to troubleshoot (after running IK) how the actual solution and the raw sensor data match, similar to how we do markers. The trouble is that when doing markers, all the data has been transformed into OpenSim frame while we don't require that for sensor data. For first round I plan to remove the unsupported option.

aymanhab:
Thanks for the review, merging based on feedback and testing by @carmichaelong and @antoinefalisse 👍

Issue #1147, opened 2019-11-07
Visualizer perspective buttons do nothing


chrisdembia:
### Steps to reproduce

1. Open GUI.
2. Open any model.
2. Click one of the cube buttons to change perspective.

### Expected result

perspective changes

### Actual result

nothing

### Environment and GUI version

OpenSim 4.1-2019-10-17-7704209

aymanhab:
@carmichaelong or @jimmyDunne can you verify in the PR build of the GUI? Thanks

PR #1152, opened 2019-11-09
include fix to visualizer toolbar buttons

aymanhab:
Fixes issue #1147 

### Brief summary of changes
Visualizer was trying to find Model object by name to compute bounding box, replaced with more robust search by uuid which is better maintained.
### Testing I've completed
Buttons performed as expected in local build

### CHANGELOG.md (choose one)

- no need to update because bug was not reproducible in public releases.


aymanhab:
@chrisdembia or @jimmyDunne If you can test the artifact that would be great. Thank you

aymanhab:
Anybody else to test so we can merge? @jenhicks @nickbianco @tkuchida

chrisdembia:
I'll test.

chrisdembia:
Works for me on Mac 10.15.

aymanhab:
Thanks @chrisdembia

Issue #1153, opened 2019-11-13
External Loads Applies Force Checkbox: Unchecking the box from the user interface is not saved


jenhicks:
### Steps to reproduce
Load gait2354 model, launch ID tool, and edit external loads by unchecking the "applies force" box
Click Save and then edit again -> applies force box is still checked
Click Save and then save the external loads file -> applies force box is still not checked when inspecting later.

### Environment and GUI version
OpenSim 4.0, Mac
Also reported by user in #1146

aymanhab:
This boils down to the following: ExternalForce uses two different/conflicting ways to indicate whether it applies_force. a custom method that checks for force_identifier and a base class method that checks for the property applies_force. Switching to using the property always will make the setting persistent and the GUI works fine but may have other side effects downstream if the API uses the custom method so I lean to not make this change but I'm open to suggestions @jimmyDunne @chrisdembia

jenhicks:
Could we remove this checkbox from the GUI? If we did that, what would be the pathway for users to change the setting?

aymanhab:
We can remove it, users will have to remove all the labels in force columns by selecting blank. which is inconvenient Another option is to keep it and make un-selection wipes out the columns for the user.  It remains the case that the API users have 2 methods appliesForce and get_applies_force() that  look identical but have different semantics/meaning.

jenhicks:
We will remove the checkboxes. We should make sure to note in the documentation and changelog that users should zero out the columns to prevent the forces from being applied.

aymanhab:
One complication is that since the radio button options (Body Force, Point Force) are not guarded by a checkbox, now users can't specify torque only, so I implemented the most surgical solution that blanks the Force columns if you uncheck the box.

jenhicks:
From @carmichaelong during beta testing (and I confirmed) ...
Tried with grf file from 2354 using the Static Optimization tool GUI. Unchecked the applies force box for 'right', and saved to file. Two issues: 1) xml file output still has applies force as true; 2) file no longer can be editted/opened from OpenSim GUI (by hitting the edit button again in the SO GUI)

aymanhab:
1) appliesForce in the file has a different meaning (used to be isDisabled before we made it 'clearer 💣 in 4.0) so it should be true even if the check-box is off as long as it applies Torque. Sad but true.
2) File fails to open because the datasource is deleted/removed while making a copy, may require fix on opensim-core. Still under investigation.

[...]

PR #1156, opened 2019-11-19
Fix external force applies force

aymanhab:
Fixes issue #1153

### Brief summary of changes
When the user unchecks the "Applies Force" checkbox, the ForceIDentifier and selected Force columns are blanked so that the internal "appliesForce" method returns false and the flag is persistent. This is the most surgical fix since it affects only rarely used scenarios without needing to change much of the functionality, documentation, or surprise users with changes to dialogs.

### CHANGELOG.md (choose one)

- updated...


sebastianskejoe:
Might also want to add similar logic when the user unchecks the applies torque checkbox. Otherwise it seems like a good fix.

aymanhab:
Build with fix available [here](https://ci.appveyor.com/project/opensim-org/opensim-gui/builds/29000434/artifacts) to verify the fix on windows. Thanks @sebastianskejoe

sebastianskejoe:
Works on Windows 10. Thanks!

aymanhab:
Thanks for testing @sebastianskejoe 👍 
Can you do a quick review @chrisdembia ?

aymanhab:
@chrisdembia this should be rather quick, thank you

aymanhab:
Thanks @chrisdembia and @sebastianskejoe

PR #1159, opened 2019-11-26
Use github actions to build installer on windows

aymanhab:
First attempt at compiling GUI in a github action

Fixes issue #0
### Brief summary of changes
replicate copy of script from cmsknee branch

### Testing I've completed
None, this only affects continuous integration
### CHANGELOG.md (choose one)

- no need to update because not user facing


aymanhab:
@chrisdembia If you can add in the line to download netbeans (since I can't get the syntax right somehow) I can pick this up since many more tools need downloading. Thank you

chrisdembia:
I took one shot. I don't know if I'll have time to keep working on this, but hopefully this gives a better starting point.

aymanhab:
@chrisdembia I now get the GUI to compile properly in netbeans but trying to figure out how to get the  latest build of opensim-core, do you know if nuget can be used or if we should setup a different mechanism? Thanks

chrisdembia:
You can refer to this:
https://github.com/opensim-org/opensim-moco/blob/4f25baf0096fee0c1288747e8f17abe866f8ef05/.github/workflows/continuous_integration.yml#L39-L67

Moco uses superbuild.

chrisdembia:
See `camsknee` branch for a functioning using GitHub Actions.

aymanhab:
Thanks @chrisdembia will definitely check your branch first.

chrisdembia:
Note that the master branch of opensim-core must be updated too; otherwise the GUI will only build with the camsknee branch of opensim-core.

aymanhab:
Thanks Chris, indeed I just wanted a starting script to work off.

aymanhab:
I'll go ahead and merge to trigger the build. We need to name the artifacts based on branch

PR #1161, opened 2020-01-07
Symlink opensim-install-command-line.sh

chrisdembia:
### Brief summary of changes

This PR causes a symlink for the `opensim-install-command-line.sh` script to be created.

### Testing I've completed

I did not build the GUI; I simply created the symlink in an existing GUI installation and ensuerd it worked.

### CHANGELOG.md (choose one)

- no need to update because...no release yet

### DIFF
--- a/Gui/opensim/build.xml
+++ b/Gui/opensim/build.xml
@@ -252,12 +252,9 @@
             <!-- '-f': overwrite if files already exist. -->
             <arg line="-f dist/OpenSim.app '${mac.dist.outer.dir}/${mac.final.app.name}'" />
         </exec>
-        <!-- copy shell script to install command-line tools to top level folder -->
-        <exec executable="mv">
-            <!-- '-f': overwrite if files already exist. -->
-            <arg line="-f '../${mac.final.app.name}/Contents/Resources/OpenSim/bin/opensim-install-command-line.sh' 
-                       '${mac.dist.outer.dir}/bin/opensim-install-command-line.sh'" />
-        </exec>
+        <!-- Move shell script to install command-line tools to top level folder -->
+        <move file="../${mac.final.app.name}/Contents/Resources/OpenSim/bin/opensim-install-command-line.sh"
+              todir="${mac.dist.outer.dir}/bin" />
         <!-- Symlinks in the outer OpenSim 4.0 folder.-->
         <symlink action="single" link="${mac.dist.outer.dir}/Geometry"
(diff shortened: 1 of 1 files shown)


aymanhab:
@chrisdembia Thanks for making the PR. It will likely work as is but did you consider "moving" the shell script instead of creating a symlink to it? Typically we create a symlink to avoid copying/exposing dependencies/libraries but that doesn't seem to be the case here. Just trying to reduce hoops if troubleshooting is needed. Thank you

chrisdembia:
You're right; a symlink is not necessary here. Moving or copying the file would work. Would you be willing to make this change?

aymanhab:
Thanks @chrisdembia I'll do it.

aymanhab:
@chrisdembia If you can review the code or test the artifact on osx that would be great since my mac environment is in limbo trying to upgrade to Catalina. Thank you.

chrisdembia:
The script inside the `.app` works now, but there is no copy of the file in `/Applications/OpenSim 4.1-.../bin`:

image

Gui/opensim/build.xml
```diff
@@ -252,7 +252,12 @@
             <!-- '-f': overwrite if files already exist. -->
             <arg line="-f dist/OpenSim.app '${mac.dist.outer.dir}/${mac.final.app.name}'" />
         </exec>
-
+        <!-- copy shell script to install command-line tools to top level folder -->
+        <exec executable="mv">
+            <!-- '-f': overwrite if files already exist. -->
+            <arg line="-f '${mac.dist.outer.dir}/bin/opensim-install-command-line.sh' 
+                       '${mac.dist.outer.dir}/bin/opensim-install-command-line.sh'" />
```
chrisdembia: This moves the file onto itself. I think the first argument should be `../${mac.final.app.name}/Contents/Resources/OpenSim/bin/opensim-install-command-line.sh`.

aymanhab: @chrisdembia I tried that and it didn't move the file either, tried the moveFile Ant command and that also failed, so I reverted to your earlier version that creates a symbolic link, we can improve later as needed. Please let me know what you think.

chrisdembia: I would prefer copying rather than moving the file, but I think a symlink is fine and we can proceed with that until a user runs into an issue.

aymanhab:
Thanks @chrisdembia

PR #1163, opened 2020-01-13
Restore preferences early

aymanhab:
Fixes issue #0

### Brief summary of changes
Restore preferences before creating the ViewDB instance since the latter uses the GeometrySearchPath Preference to locate mesh files.

### Testing I've completed
Nothing changes other than Geometry Search Oath preference showing up on OSX. 
will test artifact when ready.

### CHANGELOG.md (choose one)

- updated...


aymanhab:
This can only be tested on a machine that never had OpenSim 4.0 installed (Windows) or where defaults are cleared before launch (OSX). Things to test:
1. Geometry Search Path appears in Edit->Preferences
2. Changes to Geometry Search Path are persistent across launches.
@carmichaelong,  @jimmyDunne  @jenhicks or any other volunteers

PR #1166, opened 2020-01-16
Handle externalloads save cancel

aymanhab:
Fixes issue #1165 
### Brief summary of changes
Code that query users for file name to save external loads was not handling the cancel scenario which returns a null for filename. Check was added

### Testing I've completed
Tried save/Cancel of external loads and got no exceptions.

### CHANGELOG.md (choose one)

- not sure if this is worth mention in changelog, thoughts/feedback would be appreciated.


carmichaelong:
Just tested that it works and code LGTM.

I can go either way on whether or not to put in the changelog. The effect on users seemed minor, but sometimes it's good to document anyway if for some reason we need to refer back to when this change was made. Maybe I'd give a (very) slight preference to including in the changelog?

aymanhab:
Thanks @carmichaelong I updated the changelog. Merging.

Issue #1179, opened 2020-03-09
TypeError at "ModelbuildingFunctions.py" file


dasmehdix:
Hi, I was trying to debug the codes with scriptshell windows from [this](https://simtk-confluence.stanford.edu:8443/display/OpenSim/Scripting+in+the+GUI#ScriptingintheGUI-AvailableExampleScripts) guide.I wanted to use "ModelBuildingFunctions.py" file, I changed the comment line like below, but an error occurs. 

**The part of uncommented lines:**

`modelCopy = getCurrentModel().clone()`

`ball = addBodyToModel(modelCopy, 'ball')`

`#sphere =modeling.Sphere(0.2)`

`#attachGeometryWithOffset(modelCopy, sphere)`

`handle = connectBodyWithJoint(modelCopy, ball, 'handle', 'SliderJoint')`

`loadModel(modelCopy)`

**The Error in scriptshellwindow**
`Traceback (most recent call last):
  File "C:/Users/Eriklab2/Documents/OpenSim/4.1/Code/GUI/ModelBuildingFunctions.py", line 162, in <module>
    handle = connectBodyWithJoint(modelCopy, ball, 'handle', 'SliderJoint')
TypeError: connectBodyWithJoint() takes exactly 5 arguments (4 given)`

As far as I understand, I have to add childframe argument, but how can I add it? I am using opensim 4.1 on windows 10 OS.I am using "gait2392 model".

aymanhab:
@carmichaelong Can you reproduce?

aymanhab:
add the lines 
pelvis = modelCopy.get_BodySet().get('pelvis')
handle = connectBodyWithJoint(modelCopy, pelvis, ball, 'handle', 'SliderJoint')

carmichaelong:
@aymanhab Thanks, I was just looking at this too. To be clear, what's being added in the code above is that you're adding a "parent frame", and the ball is the child frame. Typically for these models, when you add a new body, you add the new body as a child frame and then provide the parent frame to which it's connected.

Separately, there's another issue that needs to be fixed with `ModelBuildingFunctions.py`. Currently the attach geometry line in the file provided is:
`attachGeometryWithOffset(newBody, sphere)`
But `newBody` isn't a variable. Should be `ball` instead.

aymanhab:
Thanks @carmichaelong Agreed on both, but since editing same file you can make one PR to address both issues with this out of date script

aymanhab:
[...]

PR #1180, opened 2020-03-09
fix ModelBuildingFunctions.py

carmichaelong:
Fixes issue #1179 

### Brief summary of changes
Update few lines get example working. Also rework the comments a little bit for clarity.

### Testing I've completed
Tested locally with arm26, and also changing the input string to 'r_ulna_radius_hand'

### CHANGELOG.md
Didn't update since it's so small, but it could make sense to update? Let me know either way.

### DIFF
--- a/Gui/opensim/Scripts/ModelBuildingFunctions.py
+++ b/Gui/opensim/Scripts/ModelBuildingFunctions.py
@@ -161,8 +161,8 @@ def addMuscleToModel(model, muscleName, muscleType):
    the copied model is loaded into the GUI"""
 #   modelCopy = getCurrentModel().clone()
 #   ball = addBodyToModel(modelCopy, 'ball')
-#   sphere = modeling.Sphere(0.15)
-#   attachGeometryWithOffset(ball, sphere)
+#   sphereGeometry = modeling.Sphere(0.15)
+#   attachGeometryWithOffset(ball, sphereGeometry)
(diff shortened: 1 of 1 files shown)


Gui/opensim/Scripts/ModelBuildingFunctions.py
```diff
@@ -153,12 +153,16 @@ def addMuscleToModel(model, muscleName, muscleType):
...
+   and connects the new body to a parent body called "hand". Then,
+   the copied model is loaded into the GUI"""
 #   modelCopy = getCurrentModel().clone()
 #   ball = addBodyToModel(modelCopy, 'ball')
 #   sphere = modeling.Sphere(0.15)
-#   attachGeometryWithOffset(newBody, sphere)
-#   handle = connectBodyWithJoint(modelCopy, ball, 'handle', 'SliderJoint')
-#    loadModel(modelCopy)
-
+#   attachGeometryWithOffset(ball, sphere)
```
tkuchida: "Attach sphere to ball" is a bit enigmatic; renaming `sphere` to `sphereGeometry` or similar would be clearer IMHO.

carmichaelong: Thanks, Tom, I agree that your naming scheme is much better. Should be addressed in new commit.

aymanhab:
failures are unrelated. Merging, Thanks @carmichaelong @tkuchida

Issue #1182, opened 2020-03-29
Scaling a model loses custom geometry/meshes


aymanhab:
@aymanhab commented on [Thu Feb 20 2020](https://github.com/opensim-org/opensim-core/issues/2681)

Reported on forum and by many local users in the past
https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=11542&p=32299&start=0&view=

---

@aymanhab commented on [Fri Feb 21 2020](https://github.com/opensim-org/opensim-core/issues/2681#issuecomment-589772167)

One easy fix that works is to create the model, give it a file/path at the same folder as the starting model. The downside is that if the user saves the model in a different folder the problem will re-occur.  but that's the same as any model with custom geometry

---

@aymanhab commented on [Fri Feb 21 2020](https://github.com/opensim-org/opensim-core/issues/2681#issuecomment-589773213)

@jenhicks @tkuchida @carmichaelong @chrisdembia what are your thoughts? Do you typically save the model manually? Specify a destination model file name in the setup file?

---

@aymanhab commented on [Fri Feb 21 2020](https://github.com/opensim-org/opensim-core/issues/2681#issuecomment-589773808)

Anybody else using Scaling workflow that would be interested in this discussion?

---

@aymanhab commented on [Thu Mar 26 2020](https://github.com/opensim-org/opensim-core/issues/2681#issuecomment-604658386)

This came up a couple more times during the Match 20 workshop, so higher priority. Thoughts @jenhicks 

---

@jenhicks commented on [Thu Mar 26 2020](https://github.com/opensim-org/opensim-core/issues/2681#issuecomment-604726718)

I will put it in the sprint backlog for now so we can discuss how to prioritize.

jenhicks:
@aymanhab Can you make sure there is an update in the changelog?

aymanhab:
@jenhicks Yes, changelog has been updated. Feel free to close.

Issue #1078, opened 2018-11-09
Geometry is incorrectly displayed after using Scale Tool when using a model that has its own Geometry folder


carmichaelong:

carmichaelong:
@chrisdembia @aymanhab @jimmyDunne 

Although the files in the email are not compatible with 3.3, I was able to reproduce with the files from the Rajagopal model SimTK page:
![image](https://user-images.githubusercontent.com/7455790/48288653-bbc36a80-e421-11e8-8294-442b4554adc8.png)

Note that in order to reproduce, I deleted the `r_tibia.vtp` file from the OpenSim 3.3 Geometry folder (apparently it was available in this one).

Another thing to note here is that missing Geometry may only be one problem here. If a user has their own custom geometry with the same name, then when the model loads, they won't know if their custom Geometry is used or the one in the distribution. This could lead to someone incorrectly adjusting their model, for instance, if they see "incorrect" muscle paths (e.g., muscles crossing through bones or not connecting to bones) simply due to different geometry being used.

carmichaelong:
In 4.0 using the files from the email, I got a similar effect as 3.3. Maybe I can't reproduce the same exact issue but a similar one?
![image](https://user-images.githubusercontent.com/7455790/48289264-b1a26b80-e423-11e8-85af-05a3f2c9249e.png)

chrisdembia:
@rwj1 how do these fibulas compare to those in the output of your scaling?

aymanhab:
At some point I think @jimmyDunne renamed bones since differnet models had different convention e.g. boneWhatever.vtp was meant as boneWhatever_r.vtp

carmichaelong:
Since clearly 4.0 could also find my 3.3 folder, I deleted 3.3 and tried it again in 4.0 and get something closer to what was reported:
![image](https://user-images.githubusercontent.com/7455790/48289451-68065080-e424-11e8-8bff-f25c477be63b.png)

Saving out the scaled model and loading it again yields some bones back:
![image](https://user-images.githubusercontent.com/7455790/48289537-b3b8fa00-e424-11e8-814b-ae898eed8dd9.png)

Perhaps this means that the local Geometry folder is only used on loading a model from file and not when a Tool outputs a new model?

aymanhab:
Yes, @carmichaelong that was the conjecture. If you save to the same folder as the model you started from then this whole issue goes away on reloading.

chrisdembia:
[...]

PR #1183, opened 2020-03-29
Associate scaled model with file in same folder as input model 

aymanhab:
Fixes issue #1182 

### Brief summary of changes
Associate the scaled model with a file (_scaled.osim) in the same folder as original model so that custom geometry is located using same mechanism as original model.

### Testing I've completed
Loaded Wu shoulder model and scaled it (manually) and got expected visualization.

### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/Gui/opensim/tracking/src/org/opensim/tracking/ScaleToolModel.java
+++ b/Gui/opensim/tracking/src/org/opensim/tracking/ScaleToolModel.java
@@ -313,7 +313,8 @@ public class ScaleToolModel extends Observable implements Observer {
          // Crash here
          processedModel = new Model(unscaledModel);
          processedModel.setName(scaleTool.getName());
-         processedModel.setInputFileName("");
+         String proposedFilename = unscaledModel.getInputFileName().replace(".osim", "_scaled.osim");
+         processedModel.setInputFileName(proposedFilename);
          processedModel.setOriginalModelPathFromModel(unscaledModel); // important to keep track of the original path so bone loading works
          //processedModel.setup();
 


aymanhab:
Build failure on osx is unrelated (jdk version issue). Things to test:
- Scale models with custom geometry (e.g. Wu shoulder or Rajagopal)
- Check what happens if assumed file name (originalmodel)_scaled.osim already exists.
- Try model with ContactMesh since these used to have even bigger problems with scaling

aymanhab:
scaled model is not saved on its own, the new suggested name is good, but if file exists, it's overwritten without warning which is unrelated issue. IF model is saved in different folder problem recurs. scaled model with contact mesh without issue. Ready for review/testing @carmichaelong

carmichaelong:
I like that it's a small fix, but I'm trying to understand how this works exactly. Will this change some working directory downstream by changing the inputfilename? If so, how much more code would need to be in place to separately change the working directory without making a placeholder file name?

aymanhab:
@carmichaelong This doesn't change any working directory, it just tells the scaled model that its current filename is "originalmodelpath/model_scaled.osim", this affects the behavior of the code but not the file system at all. When user decides to save scaled model, it's saved to the suggested file, if user wants to pick different file name they can choose saveAs. If they change folder they'll run into problem downstream because Geometry will not be found but that's unrelated to GUI opertion.

carmichaelong:
Tested with OpenSim-4.2-2020-03-29-4fedac9a on Win10 Pro.

Confirmed scaling Rajagopal model on 4.1 distribution led to geometry issue. Artifact here does not have that issue. This PR LGTM. Might be good to have another more general issue about how to best handle Geometry files/folder.

aymanhab:
Thanks @carmichaelong 👍 
I think issue #1078 is a duplicate or related as well. Feel free to create an issue to start discussion on how to best handle this, ideally model is a zip file with embedded meshes but that comes with a price.

carmichaelong:
That issue looks good, thanks for digging that up. Feel free to merge.

aymanhab:
Thanks @carmichaelong much appreciated. Merging

PR #1184, opened 2020-03-30
[WIP] Try xcode9.3 as claimed to use java 1.8 by default (https://docs.trav…

aymanhab:
…is-ci.com/user/reference/osx/#OS-X-Version)

Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: java
-
+jdk:
+  - oraclejdk8
 # Use container-based infrastructure to allow caching (for ccache).
 sudo: false 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - oraclejdk8
+  - openjdk8
 # Use container-based infrastructure to allow caching (for ccache).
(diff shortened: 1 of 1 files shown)


aymanhab:
@chrisdembia Per the travis page here https://docs.travis-ci.com/user/reference/osx using xcode9.3 would get java 1.8 by default, however the build fails due to trying to use open-jdk 13, any idea what we're doing wrong here? Thanks

chrisdembia:
Well, it looks like Travis-CI is failing because it can't find Ant, and GitHub Actions is failing because `brew install ant` is not working (probably temporarily).

aymanhab:
@chrisdembia True but not finding ant is a downstream effect of not finding the right java, per the page I linked to above their osx, xcode combination defaults to xcode 9.4.1 which is java 13.  Check log of last successful build shows Java 1.8 installed first

chrisdembia:
Hmmm...maybe we could try specifying the `jdk` key.

https://docs.travis-ci.com/user/languages/java/#testing-against-multiple-jdks

PR #1190, opened 2020-04-17
Upgrade swig generated files

aymanhab:
Fixes issue #0

### Brief summary of changes
Ran SWIG, good to see the differences as seen on Java/Matlab side

### Testing I've completed
Built GUI
### CHANGELOG.md (choose one)


aymanhab:
Thanks @chrisdembia

PR #1197, opened 2020-05-04
Checkout and build opensim-core for github Action

aymanhab:
Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,10 +1,14 @@
 name: opensim-application
 
 on:
+  pull_request:
+    branches:
+      - '*'
   push:
     branches: 
-      - 'action_build_core'
-      - 'master'
+      - master
+    tags:
+      - '*'
 
 jobs:
   windows:
@@ -41,7 +45,7 @@ jobs:
         cmake $env:GITHUB_WORKSPACE/opensim-core/dependencies -G"Visual Studio 16 2019" -A x64 -DSUPERBUILD_ezc3d:BOOL=on -DCMAKE_INSTALL_PREFIX=~/opensim_dependencies_install
         $env:CXXFLAGS = ""
         cmake . -LAH
-        cmake --build . --config Release -- /maxcpucount:4 
+        cmake --build . --config Release -- /maxcpucount:2 
 
     - name: Build opensim-core
       run: |
@@ -51,15 +55,14 @@ jobs:
         $env:CXXFLAGS = "/W0"
         cmake $env:GITHUB_WORKSPACE/opensim-core -G"Visual Studio 16 2019" -A x64 -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install -DBUILD_JAVA_WRAPPING=on -DBUILD_PYTHON_WRAPPING=on -DOPENSIM_C3D_PARSER=ezc3d -DBUILD_TESTING=off -DCMAKE_INSTALL_PREFIX=~/opensim-core-install
         cmake . -LAH
-        cmake --build . --config Release -- /maxcpucount:8
+        cmake --build . --config Release -- /maxcpucount:2
         
     - name: install opensim-core
       run: |
-        chdir $env:GITHUB_WORKSPACE\\build_core
         cmake --install . 
 
     - name: Checkout opensim-gui
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
 
     - name: Set up JDK 1.8
       uses: actions/setup-java@v1
@@ -119,7 +122,7 @@ jobs:
         # -q: quiet
         unzip -q opensim-core-mac.zip -d ~
     - name: Checkout opensim-gui
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
 
     - name: Set up JDK 1.8
       uses: actions/setup-java@v1
@@ -167,7 +170,7 @@ jobs:
     steps:
 
     - name: Checkout opensim-gui
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
 
     
     - name: Checkout opensim-core master
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -176,20 +176,18 @@ jobs:
     
     - name: Build opensim-core-dependencies
       run: |
-        echo $env:GITHUB_WORKSPACE\\build_deps
-        mkdir $env:GITHUB_WORKSPACE\\build_deps
+        mkdir build_deps
         cd build_deps
         $env:CXXFLAGS = "/W0"
-        cmake $env:GITHUB_WORKSPACE/opensim-core/dependencies -A x64 -DSUPERBUILD_ezc3d:BOOL=on -DCMAKE_INSTALL_PREFIX=~/opensim_dependencies_install
+        cmake opensim-core/dependencies -A x64 -DSUPERBUILD_ezc3d:BOOL=on -DCMAKE_INSTALL_PREFIX=~/opensim_dependencies_install
         $env:CXXFLAGS = ""
         cmake . -LAH
         cmake --build . --config Release -- /maxcpucount:4 
 
     - name: Build opensim-core
       run: |
-        echo $env:GITHUB_WORKSPACE\\build_core
-        mkdir $env:GITHUB_WORKSPACE\\build_core
-        chdir $env:GITHUB_WORKSPACE\\build_core
+        mkdir build_core
+        cd build_core
         $env:CXXFLAGS = "/W0"
         cmake $env:GITHUB_WORKSPACE/opensim-core -A x64 -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install -DOPENSIM_JAVA_WRAPPING=on -DOPENSIM_PYTHON_WRAPPING=on -DOPENSIM_C3D_PARSER=ezc3d
         cmake . -LAH
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -46,7 +46,7 @@ jobs:
         mkdir $env:GITHUB_WORKSPACE\\build_core
         chdir $env:GITHUB_WORKSPACE\\build_core
         $env:CXXFLAGS = "/W0"
-        cmake $env:GITHUB_WORKSPACE/opensim-core -G"Visual Studio 16 2019" -A x64 -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install -DOPENSIM_JAVA_WRAPPING=on -DOPENSIM_PYTHON_WRAPPING=on -DOPENSIM_C3D_PARSER=ezc3d -DCMAKE_INSTALL_PREFIX=~/opensim-core
+        cmake $env:GITHUB_WORKSPACE/opensim-core -G"Visual Studio 16 2019" -A x64 -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install -DBUILD_JAVA_WRAPPING=on -DBUILD_PYTHON_WRAPPING=on -DOPENSIM_C3D_PARSER=ezc3d -DCMAKE_INSTALL_PREFIX=~/opensim-core
         cmake . -LAH
         cmake --build . --config Release -- /maxcpucount:8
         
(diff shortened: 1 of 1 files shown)


aymanhab:
@chrisdembia I can checkout and build opensim-core in github action on windows but the build gets stuck, any ideas?

aymanhab:
@halleysfifthinc would you be able to help get GitHub Actions to work on linux? Thanks much

halleysfifthinc:
I will definitely take a quick look, but I won't have much/any time to spend on this till August.

.github/workflows/continuous-integration.yml
```diff
@@ -1,6 +1,9 @@
 name: opensim-application
 
-on: [push]
+on:
+  push:
+    branches: 
+      - 'action_build_core'
```
halleysfifthinc, line 6: I believe the previous behavior ran on any push (or possibly only to the master branch); possibly a good idea to also have it triggered for pushes to master?
```suggestion
      - 'action_build_core'
      - 'master'
```

.github/workflows/continuous-integration.yml
```diff
@@ -9,11 +12,48 @@ jobs:
     runs-on: windows-latest
 
     steps:
-    - name: Install opensim-core
+    - uses: actions/checkout@v1
```
halleysfifthinc, line 15: Might as well update this and other occurrences of `actions/checkout@v1`?
```suggestion
    - uses: actions/checkout@v2
```

.github/workflows/continuous-integration.yml
```diff
@@ -121,15 +161,39 @@ jobs:
...
+    - name: Checkout opensim-core master
+      uses: actions/checkout@v2
+      with:
+         name: opensim-org/opensim-core
+         repository: opensim-org/opensim-core
+         path: 'opensim-core'
+    
+    - name: Build opensim-core-dependencies
+      run: |
+        echo $env:GITHUB_WORKSPACE\\build_deps
```
halleysfifthinc, line 178: Commands run from the GITHUB_WORKSPACE folder by default; I also don't believe the `env:` is necessary, `$GITHUB_WORKSPACE` should work.
Also, the path separator on Linux is forward-slash `/`.

.github/workflows/continuous-integration.yml
```diff
@@ -121,15 +161,39 @@ jobs:
...
+        mkdir $env:GITHUB_WORKSPACE\\build_deps
+        chdir $env:GITHUB_WORKSPACE\\build_deps
+        $env:CXXFLAGS = "/W0"
```
halleysfifthinc, line 181: Environment variables can be set in the Actions yml config; see [Using Environment Variables](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables)

.github/workflows/continuous-integration.yml
```diff
@@ -121,15 +161,39 @@ jobs:
...
+      with:
+         name: opensim-org/opensim-core
+         repository: opensim-org/opensim-core
+         path: 'opensim-core'
+    
+    - name: Build opensim-core-dependencies
+      run: |
+        echo $env:GITHUB_WORKSPACE\\build_deps
+        mkdir $env:GITHUB_WORKSPACE\\build_deps
+        chdir $env:GITHUB_WORKSPACE\\build_deps
```
halleysfifthinc, line 180:
```suggestion
        cd build_deps
```

aymanhab:
Windows work now, merging while keeping branch to fix linux

aymanhab:
@chrisdembia windows now work in github actions (need caching for performance) but linux fails due to simbody-visualizer not found, any clues?

marnunez:
> @chrisdembia windows now work in github actions (need caching for performance) but linux fails due to simbody-visualizer not found, any clues?


I've encountered this before: there is some discrepancy between building the GUI against the latest opensim-core bundle from Sourceforge and building it against opensim-core master. I belive the SDK folder is not created automatically in opensim-core unless specified in CMake. Could it be that? There's something the old linux build does right that the newer doesn't
By the way: solving this would also solve opensim-gui being built against and old Python 3.4 release. Nice.

halleysfifthinc:
opensim-core needs to be built with `-DOPENSIM_INSTALL_UNIX_FHS=OFF` for the GUI. This is currently the case for the deployed builds of opensim-core (based on [.travis.yml](https://github.com/opensim-org/opensim-core/blob/master/.travis.yml)). On *nixes, this defaults to `ON`, so if you build opensim-core yourself, you will need to specify the alternative when you run cmake.

marnunez:
> opensim-core needs to be built with `-DOPENSIM_INSTALL_UNIX_FHS=OFF` for the GUI. This is currently the case for the deployed builds of opensim-core (based on [.travis.yml](https://github.com/opensim-org/opensim-core/blob/a470a0e7748997f87239cad4dd98eee1d2a7bb2a/.travis.yml#L218)). On *nixes, this defaults to `ON`, so if you build opensim-core yourself, you will need to specify the alternative when you run cmake.

:facepalm:

aymanhab:
Thanks for the pointer @marnunez and @halleysfifthinc 👍 

I made the change to see if it goes thru and we get an artifact but long term we should make the CMake defaults work out of the box i.e. make GUI build assume OPENSIM_INSTALL_UNIX_FHS=ON or change the flag to OFF by default if it has no other downstream side-effects.

chrisdembia:
> we should make the CMake defaults work out of the box i.e. make GUI build assume OPENSIM_INSTALL_UNIX_FHS=ON or change the flag to OFF by default if it has no other downstream side-effects.

I think the correct approach is that the GUI's CMakeLists.txt should ensure opensim-core was built with `OPENSIM_INSTALL_UNIX_FHS=OFF`. I think it's useful to have `OPENSIM_INSTALL_UNIX_FHS=ON` as the default in opensim-core, because Linux users would be surprised by an installation layout that does not conform to the Filesystem Hierarchy Standard, especially because users may install opensim-core system-wide.

aymanhab:
Thanks for chiming in @chrisdembia but the GUI does not "build" opensim-core internally, instead it assumes it was already built. Can you explain the rationale for building the GUI against the non-preferred layout on linux, rather than modifying the build script?

chrisdembia:
> Thanks for chiming in @chrisdembia but the GUI does not "build" opensim-core internally, instead it assumes it was already built.

I understand. I meant to say that the GUI can raise an error if it detects that opensim-core was built with `OPENSIM_INSTALL_UNIX_FHS=ON`.

> Can you explain the rationale for building the GUI against the non-preferred layout on linux, rather than modifying the build script?

When you and I discussed the layout of the GUI on Mac, we confronted this same issue. I originally thought it would be better to use the UNIX FHS layout for the GUI on Mac, but I think we mutually arrived at the conclusion that using the same layout of the GUI across platforms was the right way to go.

aymanhab:
Thanks @chrisdembia I opened issue https://github.com/opensim-org/opensim-gui/issues/1200 to continue the discussion of desired behavior.

aymanhab:
With artifacts from windows and linux available here https://github.com/opensim-org/opensim-gui/actions/runs/114987515 this PR is ready for review. We can try adding caching for opensim-core now or do it later when commits on this repo are more frequent. Would be great if somebody on ubuntu can test the artifact as well.

.github/workflows/continuous-integration.yml
```diff
@@ -1,6 +1,10 @@
 name: opensim-application
 
-on: [push]
+on:
+  push:
+    branches: 
+      - 'action_build_core'
+      - 'master'
```
chrisdembia, line 7: I suggest we use what we do for Moco:

```
on:
  pull_request:
    branches:
      - '*'
  push:
    branches: 
      - master
    tags:
      - '*'
```

.github/workflows/continuous-integration.yml
```diff
@@ -9,11 +13,51 @@ jobs:
...
+
+    - name: Build opensim-core
+      run: |
+        echo $env:GITHUB_WORKSPACE\\build_core
+        mkdir $env:GITHUB_WORKSPACE\\build_core
+        chdir $env:GITHUB_WORKSPACE\\build_core
+        $env:CXXFLAGS = "/W0"
+        cmake $env:GITHUB_WORKSPACE/opensim-core -G"Visual Studio 16 2019" -A x64 -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install -DBUILD_JAVA_WRAPPING=on -DBUILD_PYTHON_WRAPPING=on -DOPENSIM_C3D_PARSER=ezc3d -DBUILD_TESTING=off -DCMAKE_INSTALL_PREFIX=~/opensim-core-install
+        cmake . -LAH
+        cmake --build . --config Release -- /maxcpucount:8
```
chrisdembia, line 54: Any reason for using 8 here and 4 above?

.github/workflows/continuous-integration.yml
```diff
@@ -9,11 +13,51 @@ jobs:
...
+    - name: install opensim-core
       run: |
-        (New-Object System.Net.WebClient).DownloadFile("https://prdownloads.sourceforge.net/myosin/github-actions/opensim-core/opensim-core-win.zip", "opensim-core-win.zip")
-        7z x opensim-core-win.zip -oC:\\
-        dir C:\\opensim-core
+        chdir $env:GITHUB_WORKSPACE\\build_core
+        cmake --install . 
+
     - name: Checkout opensim-gui
       uses: actions/checkout@v1
```
chrisdembia, line 62: This is redundant with the first step above.

.github/workflows/continuous-integration.yml
```diff
@@ -121,15 +165,38 @@ jobs:
...
 
     steps:
-    - name: Install opensim-core
-      run: |
-        wget https://prdownloads.sourceforge.net/myosin/opensim-core/opensim-core-latest_linux_Release.zip
-        # -q: quiet
-        unzip -q opensim-core-latest_linux_Release.zip -d ~
 
     - name: Checkout opensim-gui
       uses: actions/checkout@v1
```
chrisdembia, line 170: There are inconsistencies between platforms related to the step for checking out `opensim-gui`.

.github/workflows/continuous-integration.yml
```diff
@@ -121,15 +165,38 @@ jobs:
...
+    - name: Install packages
+      run: sudo apt-get install --yes liblapack-dev freeglut3-dev libxi-dev libxmu-dev doxygen python3 python3-dev python3-numpy python3-setuptools swig
+
+    - name: Build opensim-core-dependencies
+      run: |
+        mkdir build_deps
+        cd build_deps
+        cmake ../opensim-core/dependencies -DSUPERBUILD_ezc3d:BOOL=on -DCMAKE_INSTALL_PREFIX=~/opensim_dependencies_install
+        cmake . -LAH
+        cmake --build . --config Release
```
chrisdembia, line 189: I'm not sure `--config` has meaning for generators other than MSVC and Xcode. It's more typical to put `-DCMAKE_BUILD_TYPE=Release` when configuring.

.github/workflows/continuous-integration.yml
halleysfifthinc: FYI The Actions CI runners only have 2 cores.

aymanhab: Thanks @chrisdembia and @halleysfifthinc changed to 2

.github/workflows/continuous-integration.yml
aymanhab: Done

.github/workflows/continuous-integration.yml
aymanhab: Now all use checkout@v2

.github/workflows/continuous-integration.yml
aymanhab: Left alone since it works as is and these experiments are expensive.

aymanhab:
Thanks for the feedback @chrisdembia I addressed it, ready for re-review.

.github/workflows/continuous-integration.yml
```diff
@@ -9,13 +17,53 @@ jobs:
...
       run: |
-        (New-Object System.Net.WebClient).DownloadFile("https://prdownloads.sourceforge.net/myosin/github-actions/opensim-core/opensim-core-win.zip", "opensim-core-win.zip")
-        7z x opensim-core-win.zip -oC:\\
-        dir C:\\opensim-core
+        chdir $env:GITHUB_WORKSPACE\\build_core
+        cmake --install . 
+
     - name: Checkout opensim-gui
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
```
chrisdembia, line 66: The GUI was already checked out in the first step; no need to check out again.

.github/workflows/continuous-integration.yml
```diff
@@ -75,7 +123,7 @@ jobs:
         # -q: quiet
         unzip -q opensim-core-mac.zip -d ~
     - name: Checkout opensim-gui
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
```
chrisdembia, line 126: Can the location of the step that checks out the GUI be consistent between all platforms?

aymanhab:
I addressed your latest feedback, once windows and linux build successful will merge. I didn't work on OSX, will do later following the linux workflow mostly. Thanks for the review and feedback @chrisdembia

chrisdembia:
LGTM.

We should fix the Mac build as soon as we can, because the Travis Mac build is also failing.

aymanhab:
Thanks much @chrisdembia Merging

PR #1202, opened 2020-05-25
Added Debian bullseye support for the installer

marnunez:
### Brief summary of changes
Changed to use ID of OS instead of NAME
Only use VERSION_ID if ID == "ubuntu"
### Testing I've completed
Built on my own Debian bullseye/sid


aymanhab:
Thanks for submitting this @marnunez, quick feedback:
1. whitespace/tab changes makes it hard to see the actual diffs, can you fix that?
2. In our current work-from-home environment, and moving forward it is hard to access/test linux installers, so it would be good to leverage/test the artifacts produced by github actions on actual machines running linux.

marnunez:
1) Sorry about the autoformatting, my bad.
2) I see what you mean, how would you suggest to approach this? There's no ready github actions build for a Debian environment, but something more fancy using self-hosted runners might be done if needed. As for now, I did build and install this PR with the Debian machine I'm using, for what it's worth...

aymanhab:
@marnunez If you fix the formatting, and @halleysfifthinc approves the change (to make sure it doesn't break his changes, I don't have access to either platforms) that would be great.  Also please include in this PR, adding your name to the list of contributors in the contributing.md file. Thank you

Gui/opensim/installer_files/Linux/INSTALL
```diff
@@ -114,27 +119,27 @@ if [[ $OS == "Ubuntu" ]] && [[ $VER == "18.04" ]]; then
...
-    else
-        ./opensim-install-command-line.sh > /dev/null
-    fi
-else
-    err_unsupported_env
+    Version=1.0
+    Type=Application
+    Name=OpenSim
+    Comment=OpenSim is an open source software for neuromusculoskeletal modeling, simulation \
+        and analysis.
```
halleysfifthinc, line 126:
```suggestion
    Comment=Neuromusculoskeletal modeling, simulation, and analysis.
```
We had changed this in the standalone desktop file but I noticed that I had forgot to make that change here as well.

Gui/opensim/installer_files/Linux/INSTALL
```diff
@@ -83,26 +83,31 @@ done
...
+    if ! [ -z "${PRETTY_NAME+x}" ]; then
+        OS_FULL=$PRETTY_NAME
+    fi
 else
     err_unsupported_env
-fi
-
-if [[ $OS == "Ubuntu" ]] && [[ $VER == "18.04" ]]; then
+fi 
+if ([[ $OS == "ubuntu" ]] && [[ $VER == "18.04" ]]) || ([[ $OS == "debian" ]] && [[ $OS_FULL =~ "bullseye" ]]); then
```
halleysfifthinc, line 98: Is there a reason (ie in terms of an expectation of a non-exact match) for using the `=~` comparison for matching "bullseye"? I'm not familiar with Debian

Gui/opensim/installer_files/Linux/INSTALL
```diff
@@ -83,26 +83,31 @@ done
...
 if [[ -e /etc/os-release ]]; then
-    . /etc/os-release
-    OS=$NAME
-    VER=$VERSION_ID
+    . /etc/os-release 
+    OS=$ID
+    OS_NAME=$NAME
+    if ! [ -z "${VERSION_ID+x}" ]; then
+        VER=$VERSION_ID
+    fi
```
halleysfifthinc, line 91: My bash skills aren't the greatest (and I wasn't able to figure it out with some quick googling); can you explain what the `+x` is doing? Also, if you're negating the `-z` option, wouldn't using `-n` or even just `[ "$VERSION_ID+x}" ]` be more concise/readable?

Gui/opensim/installer_files/Linux/INSTALL
```diff
@@ -83,26 +83,31 @@ done
 ## End from https://stackoverflow.com/a/29754866
 
 if [[ -e /etc/os-release ]]; then
-    . /etc/os-release
-    OS=$NAME
-    VER=$VERSION_ID
+    . /etc/os-release 
+    OS=$ID
+    OS_NAME=$NAME
```
halleysfifthinc, line 88: This variable appears to be unused now and can be removed?

halleysfifthinc:
Regarding the Actions CI artifacts, I believe what @aymanhab is suggesting is to check that the installer produced by CI works by test installing it on your machine.

PR #1203, opened 2020-05-29
GitHub Actions: Install netbeans on Mac

chrisdembia:
### Brief summary of changes

- Fix installation of NetBeans on Mac.
- Miscellaneous cleanup and organization.

### CHANGELOG.md (choose one)

- no need to update because...not user-facing

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -144,7 +144,7 @@ jobs:
         echo "TODO2"
         wget https://download.netbeans.org/netbeans/8.2/final/bundles/netbeans-8.2-javase-macosx.dmg
         hdiutil attach netbeans-8.2-javase-macosx.dmg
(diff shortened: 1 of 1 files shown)


chrisdembia:
@aymanhab I think the issue is that NetBeans is finding JDK 14, not JDK 8.

chrisdembia:
Ready for review.

aymanhab:
@chrisdembia Do you think now is a good time to stop building on travis so we have one place to maintain?

chrisdembia:
> @chrisdembia Do you think now is a good time to stop building on travis so we have one place to maintain?

As soon as the builds are green for all three OSes, then yes. It seems like this is indeed the case.

chrisdembia:
@aymanhab I've gone ahead and turned off the Travis-CI and AppVeyor continuous integration. I suggest we leave the related CI scripts in the repo in case we want to fall back on these services.

aymanhab:
Merged, thanks @chrisdembia

Issue #1204, opened 2020-06-04
Mac: GUI does not start


chrisdembia:
### Steps to reproduce

Launch the GUI

### Expected result

GUI launches

### Actual result

<img width="532" alt="image" src="https://user-images.githubusercontent.com/846001/83821928-8c6bf280-a684-11ea-954f-fc0e7c4a509b.png">

I think the issue is that an exception is thrown if the logging file `opensim.log` can't opened, because Macs use `/` as the working directory for applications.

### Environment and GUI version

Mac 4.2-2020-06-02-7baafaf

aymanhab:
@chrisdembia if you sudo launch do you get the file and no exception?  How do you think that worked in the past with out.log or err.log?

chrisdembia:
I'm not willing to try that on my machine; I could try it in a virtual machine. I imagine we just were more lenient about logging to a file. 

Maybe, if we cannot log to the current directory, we should log to a file like `~/.opensim/opensim.log`.

aymanhab:
I understand @chrisdembia but where do the err.log and out.log file get written in the curren GUI that you locally build, or the published 4.1? I can try that myself as well but thought you have this info already. Let me know either way to proceed. Thank you

chrisdembia:
As far as I know, they are not written (because we try to write them to `/` and we can't).

aymanhab:
@chrisdembia when I try to launch the GUI on osx by browsing to the bin folder it actually launches without an issue, is that consistent with your conjecture? Do you know where the messages.log file are typically stored on osx as I couldn't locate them. Thank you

chrisdembia:
The `opensim.log` file is written to your current working directory.

>  is that consistent with your conjecture? 

If you installed the GUI in a user-writable directory, then yes. If not, then no.

aymanhab:
Installing for all users, the error comes from permissions of the file etc/OpenSim.conf which used to be -rw-r--r-- and instead are now -rw-------. Were we doing something special to this file when building installer on travis?

chrisdembia:
Woah how did you figure that out? On my machine, I have the same permissions issue. 

When I build the GUI locally, the permissions are `-rw-r--r--`.

chrisdembia:
Fixing the permissions of `OpenSim.conf` did not allow the GUI to start for me.

aymanhab:
[...]

PR #1208, opened 2020-06-15
Fix mac installer permissions

aymanhab:
Fixes issue #1204 

### Brief summary of changes
chmod of conf file while building installer on mac
### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -78,16 +78,16 @@ jobs:
         cmake . -LAH
         cmake --build . --config Release -- /maxcpucount:2 
 
-    - name: Cache opensim-core
-      id: cache-core
-      uses: actions/cache@v1
-      with:
-        path: ~/opensim-core-install
-        # https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows#matching-a-cache-key
-        key: ${{ runner.os }}-${{ hashFiles('opensim-core/*') }}
-
+    #- name: Cache opensim-core
+    #  id: cache-core
+    #  uses: actions/cache@v1
+    #  with:
+    #    path: ~/opensim-core-install
+    #    # https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows#matching-a-cache-key
+    #    key: ${{ runner.os }}-${{ hashFiles('opensim-core/*') }}
+    #
     - name: Build opensim-core
-      if: steps.cache-core.outputs.cache-hit != 'true'
+      # if: steps.cache-core.outputs.cache-hit != 'true'
       run: |
         echo $env:GITHUB_WORKSPACE\\build_core
         mkdir $env:GITHUB_WORKSPACE\\build_core
@@ -96,12 +96,7 @@ jobs:
         cmake $env:GITHUB_WORKSPACE/opensim-core -G"Visual Studio 16 2019" -A x64 -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install -DBUILD_JAVA_WRAPPING=on -DBUILD_PYTHON_WRAPPING=on -DOPENSIM_C3D_PARSER=ezc3d -DBUILD_TESTING=off -DCMAKE_INSTALL_PREFIX=~/opensim-core-install
         cmake . -LAH
         cmake --build . --config Release -- /maxcpucount:2
-        
-    - name: Install opensim-core
-      if: steps.cache-core.outputs.cache-hit != 'true'
-      run: |
-        chdir $env:GITHUB_WORKSPACE\\build_core
-        cmake --install . 
+        cmake --install .
 
     - name: Update submodules
       run: git submodule update --init --recursive -- opensim-models opensim-visualizer Gui/opensim/threejs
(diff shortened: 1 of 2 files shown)


aymanhab:
@chrisdembia with this PR, install for user-only works normally, install for all users works as well but the short cut is not created (have to browse down to MacOS folder under .app to launch). Any ideas?

.github/workflows/continuous-integration.yml
```diff
@@ -76,8 +78,16 @@ jobs:
         cmake . -LAH
         cmake --build . --config Release -- /maxcpucount:2 
 
+    - name: Cache opensim-core
+      id: cache-core
+      uses: actions/cache@v1
+      with:
+        path: ~/opensim-core-install
+        # https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows#matching-a-cache-key
+        key: ${{ runner.os }}-${{ hashFiles('opensim-core/*') }}
```
chrisdembia, line 87: What is the desired caching behavior? Right now, the cache is invalidated any time the master branch of opensim-core is updated. Another option is for this repo to contain a file that lists the desired commit/tag of opensim-core.

If this is the case, keep in mind that these actions should not be used to create formal releases.

chrisdembia:
> the short cut is not created

Which shortcut?

aymanhab:
The launchpad shortcut.

chrisdembia:
Oh, I don't typically use launchpad. What about double-clicking the .app using Finder?

.github/workflows/continuous-integration.yml
aymanhab: @chrisdembia It's a good point. As of now I'm using master to avoid unnecessary rebuilds when GUI changes but core doesn't, and also so that we detect when a change in core breaks the GUI build. At some point we need to switch to a specific commit for core which ideally is specified in a file/hash. Not sure yet how to set this up in github actions without modifying the yml file when we switch. Separate workflow?

chrisdembia: Thanks for clarifying; it would be nice to add a comment expressing your first point.

A few thoughts:
- We can add a plain-text file `opensim-gui/opensim-core-commit.txt` that we use to invalidate the cache for a pre-defined version of opensim-core.
- We could have a separate workflow that is a nightly build that checks to see if there is a new commit of `opensim-core`, and if so, uploads an updated artifact that the continuous-integration workflow downloads (using `actions/download-artifact`).
- When the continuous-integration script is not burdened with building opensim-core, it'll be snappier, and we could have an additional build that tests against a released version of opensim-core.

chrisdembia:
These file permission issues are really curious.

chrisdembia:
I wonder if the issue is that I use `sudo` in earlier steps in the GitHub Actions script. Perhaps we just need to add `sudo -k` to undo the sudo?

https://superuser.com/questions/554476/how-do-i-back-out-of-sudo-privileges-in-os-x

aymanhab:
Latest fix to chmod in build.xml did the trick, including icon in launchpad. We can investigate alternatives but I lean to merge so that  builds are usable on osx since travis build is broken as well.
Not sure how well caching works since core has been changing. Do you know if a rebuild invalidates the cache @chrisdembia ?

Gui/opensim/build.xml
```diff
@@ -232,6 +232,8 @@
             value="${app.version}">
             <replacefilter token="@VERSION@" />
         </replace>
+        <chmod file="${mac.dist.inner.dir}/etc/OpenSim.conf" perm="744"/>
+        <chmod file="dist/OpenSim.app/Contents/Info.plist" perm="644"/>
```
chrisdembia, line 236: Can you add some comments (perhaps describing why you're using 744 for one and 644 for another) and a TODO indicating that this is temporary?

.github/workflows/continuous-integration.yml
```diff
@@ -88,17 +98,10 @@ jobs:
         cmake --build . --config Release -- /maxcpucount:2
         
     - name: Install opensim-core
```
chrisdembia, line 100: It's not great that there is a separate Install step for one platform but not the others.

chrisdembia:
> Not sure how well caching works since core has been changing.

I'm re-running the CI jobs to see if the cache is working.

> Do you know if a rebuild invalidates the cache @chrisdembia ?

What do you mean by a rebuild?

> since travis build is broken as well.

? We've disabled travis.

chrisdembia:
@aymanhab one reason the cache may change even if the opensim-core master branch is not updated: the `.git` folder will contain any new activity on *other* branches.

I think we need to use a different `key` that ignores the `.git` directory: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

My suggestion is to clone `opensim-core`, run a git command to obtain the commit hash for master, write that hash to a file, and then use `hashFiles()` on that file.

aymanhab:
I think long term using upload artifact from core and download artifact from gui would be a more sustainable solution that decouples the two repos and avoid duplication (e.g. the improvement you made on core to install order), in the interim a file with commit hash would be more reliable than what we have. 
I've been hesitant to make changes when the actions work successfully since every change takes about two hours to run, was hoping caching will help but not sure it does at the moment.

chrisdembia:
>  think long term using upload artifact from core and download artifact from gui would be a more sustainable solution that decouples the two repos and avoid duplication (e.g. the improvement you made on core to install order), in the interim a file with commit hash would be more reliable than what we have.

Unfortunately, I do not think this is possible: downloading an artifact from a different repository.

My suggestion was to use a different workflow in this repository to upload the artifact. Actually, it seems that using artifacts across workflows is not even possible yet... https://github.com/actions/upload-artifact/issues/21

https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts

https://github.community/t/api-to-download-artifact/16496/12

chrisdembia:
> I've been hesitant to make changes when the actions work successfully since every change takes about two hours to run, was hoping caching will help but not sure it does at the moment.

I suggest commenting out the caching of `opensim-core` for now, merging this, and then we can address the caching separately.

aymanhab:
Sounds reasonable, thanks @chrisdembia

.github/workflows/continuous-integration.yml
aymanhab: Removed. Thanks @chrisdembia

aymanhab:
Changes done as suggested, actions built successfully. Ready for review

chrisdembia:
LGTM.

aymanhab:
Thanks @chrisdembia

PR #1210, opened 2020-06-19
Update continuous-integration.yml

aymanhab:
exit sudo mode after installing tools on osx

Fixes issue #1204 

### Brief summary of changes
exit sudo in workflow, remove extra chmods from build.xml

### CHANGELOG.md (choose one)

- updated...


aymanhab:
Resulting artifact after undoing sudo and permissions has the same old problems and doesn't launch. Any other ideas to try, @chrisdembia ? Thanks

Issue #1188, opened 2020-04-09
Provide access to IMU-Based IK in the GUI


jenhicks:
@jenhicks commented on [Thu Apr 09 2020](https://github.com/opensim-org/opensim-core/issues/2720)

aymanhab:
Here's a mock UI for IMU based IK. Main issues to address/discuss:

1. Whether have separate Calibration Tool (opens door for custom algorithms/options) or fold into IK, or make optional.
2. What options to expose: Example weights? Error Reporting?
![CaptureIMU_IKv1](https://user-images.githubusercontent.com/3972208/92286187-25e0b580-eebb-11ea-8d00-915cbb9b6978.JPG)

aymanhab:
Thoughts? @jenhicks @carmichaelong @nickbianco

carmichaelong:
Thanks for mocking this up. Some responses below:
1. Seems like a separate calibration tool would be good (though even just a separate "Calibration" section even if it's on the same tab could be good)? I don't feel strongly about this until I see some other users work with it.
2. Error reporting would be good. A similar weighting menu to markers would be good so that current users would find it intuitive.

A few additional notes:
- `Rotations from IMU world...`: Are these in degrees? What's the order of rotations?
- `Label in Calibration file` is a little unclear. This is the body that is used for determining the model's "forward" direction, correct?
- `Heading Axis`: Is this the axis from the calibration file or the model's body?
- Is this assuming that "x" will always be the forward direction for the model's body to be aligned?

nickbianco:
Does the calibration step only need to happen once before running several IK trials? If so, then being separate might make sense for most workflows..

aymanhab:
Thanks for the great feedback. 
I draw the parallel to MarkerPlacer use in scaling which is used once or is optional but is shown on same dialog. To have separate dialogs and hope users specify same transform is error prone. Definitely need popup help to clarify usage.  Hoping to minimize custom ui for weights without sacrificing convenience.

aymanhab:
@aseth1 This is the starting point for the IMU based IK user interface. I look forward to your feedback.

jenhicks:
Thanks for the update @aymanhab . I agree with @carmichaelong 's comments.

You might add some more text in the interface to separate the different parts of the process. For example:
(1) Transform data from IMU world frame to OpenSim's reference frame

(2) Place IMU sensors on the OpenSim model
File with orientations of the IMUs while the subject is in the OpenSim model's default pose:
[...]

PR #1223, opened 2020-09-26
[WIP] Ik with imus

aymanhab:
Fixes issue #1188 

### CHANGELOG.md (choose one)

- updated...


aymanhab:
@carmichaelong I fixed the issue with the crash after running IK so you can try the latest build on ci. The model/data-files from the opensense test case are a good place to start or you can try the files from Johanna. Thank you

carmichaelong:
Went through and tested a few cases (some ill-formed, some not) based on the test cases. I think the main usage works, but some improper usage causes some issues:

- Without a model loaded, the Tool looks like it can be chosen from the drop-down menu. Clicking on it does not do anything.
image

- Loading `std_calibrated_subject07.osim` allows you to choose the option. It worked with two files (`MT_012005D6_009-quaternions_RHJCSwinger.sto` and `MT_012005D6_009-quaternions_calibration_trial_Facing_X.sto`).

- Trying to use with `subject07.osim` (without calibrated IMUs) does not run the analysis, but also doesn't give the user any feedback that the model is deficient for this case.

- Trying to write the output file as a relative file seems to print results to somewhere unexpected. For instance, using `out3.mot` has the output print out: `Wrote IK with IMU tracking results to: './/out3.mot'.`

- Trying to use one of the setup files (e.g., `setup_IMUInverseKinematics_HJC_trial.xml`) and running the tool crashes OpenSim. See log: 
[messages.log](https://github.com/opensim-org/opensim-gui/files/5358097/messages.log)

aymanhab:
Thanks for the feedback and testing @carmichaelong, some of these issues have been addressed in later iterations, in particular the last 2 items, so keep an eye.
I believe handling the model with no imus should be done on the core side to avoid duplication since same issue will happen running from the command line or Matlab.

PR #1225, opened 2020-10-14
Ik with imus

aymanhab:
Fixes issue #0

### Brief summary of changes
Introduce functionality to Calibrate a model for use with IMUs and to run IK to track sensor orientations. 

### Testing I've completed
Ran setup files from opensense test cases in the GUI and produced expected results/behavior.

### CHANGELOG.md (choose one)

- updated...


aymanhab:
Functional version of UI for calibration and tracking sensor data. Ready for testing @carmichaelong @jenhicks

aymanhab:
Issues from earlier testing are here https://github.com/opensim-org/opensim-gui/pull/1223#issuecomment-706448109 so that future discussion is in one place/PR.

aymanhab:
Issue with disabling commands on entry or with no model has been addressed by the latest commit.

carmichaelong:
Thanks @aymanhab. Some issues are confirmed fixed:
- IMUIK is grayed out without a model
- Using a setup file (e.g., `setup_IMUInverseKinematics_HJC_trial_nofeet.xml`) works correctly

Some issues remain:
- Using a relative path yields an odd output (i.e., using `out` as the output motion file name prints out in the console: `Wrote IK with IMU tracking results to: './/out5'.`)
- No message for using a model without IMUs (as you said, this is a TODO for core, but just want to keep this documented)

Other issues I noticed (not necessarily new):
- Running the tool without any input yields a non-descriptive error message:
image
- Output file seems to always at the `.mot` extension regardless of the current file extension. This could be good if the user enters a name that does not have an extension, but using the setup file (`setup_IMUInverseKinematics_HJC_trial_nofeet.xml`) yields a file named `ik_MT_012005D6_009-quaternions_RHJCSwinger.mot.mot`

aymanhab:
@jenhicks you can get the latest build on github Action build of this PR. Please feel free to pass along.

carmichaelong:
@aymanhab Changes on core side have fixed almost everything. Still having a weird issue with inputting a relative path (when using the dialog box from scratch).:
image

jenhicks:
@aymanhab Is there a command I can use to import the IMU data via the GUI scripting shell?

aymanhab:
@jenhicks The GUI and API tools assume all the data importing/conversion was done beforehand so you start with a .sto file with quaternions. All options for importing/converting data from the imu system/formats to OpenSense are available using the scripting shell. I tested with files from the openense test case at 
https://github.com/opensim-org/opensim-core/tree/master/Applications/opensense/test

jenhicks:
@aymanhab Can you send me the test case files?

aymanhab:
Here you go Jen.

Also available at:

https://github.com/opensim-org/opensim-core/tree/master/Applications/opensense/test

jenhicks:
I am getting an error during the calibration step when running in the GUI.

I am using the files from the test case.
I load the model subject_07
When I run the IMU calibration tool, I get an error when trying to load the orientations file. imuOrientations.sto
"expected = double received = Quaternion"

aymanhab:
@jenhicks That should work but I'll download the artifact to check. What platform are you on? and which build/artifact are you using?

jenhicks:
Mac OSX and 4.2-2020-10-16-60522f1 (

aymanhab:
Testing on windows worked flawlessly out of the box, checking Mac for platform specific issues..

aymanhab:
Reproduced on Mac osx, thanks for the find @jenhicks Will let you know as I investigate.

jenhicks:
Sounds good. As we discussed, the other thing I noticed for calibration was that the heading correction was pre-populated with some segments before I loaded the orientations file.

aymanhab:
@antoinefalisse this PR adds IK with IMUs to the OpenSim application. Please let me know if you can't locate the artifact/build to check it out. Thank you

aymanhab:
The issue on OSX is line ending/format issue since running files from source tree works as expected! Exploring options/solutions..

aymanhab:
I'm trying to add a Model, setup files to the installation to run OpenSense, this will solve the issue that @jenhicks ran into (copying files cross platform) and give easy entry point for new/interested users. Questions to answer while working on this (since it is done in the models repo):
- Name of the model (Rajagopal2015 based but with different bones and no muscles, maybe simplified) what about Rajagopal2015_opensense or Rajagopal2015_simplified.
- Should I include an OutputReference folder?
- Should we remove the duplicated models/setup files included now under OpenSenseExample?

Thoughts @carmichaelong @jenhicks

antoinefalisse:
@aymanhab I have started doing some testing, 3 comments so far

- I think that's what Jen said yesterday but there are weird default options for the choice of the base sensor. Before loading any orientation file, I have the option between `pelvis`, `shank_r`, and `calcn_r`. I think it would make more sense to have blank there when you open the tool for the first time, and nothing in the drop-down menu.

- There is no blank option for the base sensor name (there is one for the axis), which is confusing because it says: "leave blank if no heading correction is desired". Is leaving the axis choice blank sufficient?

- I have a problem when running the calibration. The error message is thrown [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/OpenSense/IMUPlacer.cpp#L90). When saving the calibration setup file, I see that there is no `<model_file></model_file>` in the saved setup xml file. If I add the path to the model manually in the xml and then load the setup file then it is ok, but it seems that the current model (in bold in the GUI) is not selected "by default".

After calibration, I could run the IMU IK without issues with some of my data.
I can give it a second pass later if needed.

aymanhab:
Thanks for testing @antoinefalisse 
- PR to fix core side you pointed out has been opened.
- Default names in dropdown of imu names have been removed, When list is populated from file, the blank option will not be available but the blank option for axis is always available and that's what we key off on the core side.

aymanhab:
The latest build/artifact has all the fixes above, and a new folder under Models/Rajagopal_OpenSense that contains all the models and data files to run calibration and IK with IMUs. I welcome feedback on the naming and model/data/setup files. The only issue left to address AFAIK is how to handle user typing relative path in output motion file specification. Please let me know if there's anything else I'm missing @jenhicks @carmichaelong @antoinefalisse @nickbianco

aymanhab:
Tested on osx and worked exactly as expected out of the box.

aymanhab:
Ready for another round of testing/review. Fix on core for missing model_file is a separate PR under review.

antoinefalisse:
@aymanhab can I review this once #2898 is merged in core? Or would you like feedback already or feedback on #2898?

aymanhab:
Thanks @antoinefalisse I'd wait for #2898 so that you can test both core changes and GUI changes in close to final form. If you want to take a look at PR #2898 which also addresses issue #2752

Issue #1137, opened 2019-09-26
Visualizer custom settings


aymanhab:
The following changes to default visualization lead to better visuals (annotated for ease of implementation, impact smaller is better)

1. Adjust the camera's field of view (particularly helpful for images in sagittal plane) (1, 1)
2. Adjust the near and far clipping planes so model is rendered at maximum resolution (ideally, use separate rendering pass for environment) (3, 1)
3. Change the color/intensity/position of the sunlight, ambient light, and camera light (1, 1)
4. Add a faint light beneath the model (color depends on floor color) to model reflection from sunlight (1, 3)
5. Change floor material to remove shininess (1, 4)
6. Change the color of the bones (3, 1)
7. Compute texture coordinates and texture map the bones (may cause GUI slowdown/crash) (4, 1)
8. Fix the marker lighting (https://github.com/opensim-org/opensim-gui/issues/1107) (1, 2)
9. Allow users to export images at much higher resolution (?, 2)
10. Increase shadow frustum size and shadow map depth to reduce rendering artifacts (3, 3)

aymanhab:
@tkuchida I don't think there's a single place that's a good fit for all items on this list:
- Object specific changes (e.g. Bone-texture) belong to model files
- Some settings e.g. related to lighting, floor, and Camera belong to the View Controls or the Toolbar
- Other settings (e.g. those that don't take effect immediately and require a restart) belong to a more permanent storage medium (e.g. files that get read on launch), building a GUI for them while not interactive in nature would be of limited use.

I'd lean to have a file with a list of "presets" that include camera parameters, lights, etc. and users have a persistent preference that selects among those presets, with a special "custom" preset that would allow expert users to replicate settings with minimum effort,rather than have a sophisticated GUI that gets used/exercised by a very slim fraction of users. Sure would love to hear your thoughts as our most expert user in this area as well.

tkuchida:
Not sure that would be the best way to split up settings. For example, if the user can specify in the model file that a mesh is metal or bone, then the GUI could provide an interface to adjust the visualization parameters for all meshes of a given material. So the "object-specific changes" would be implemented partly in the model file and partly in the GUI. Also, there are settings that currently require a restart, but it may be possible to update these live if the right things are updated programmatically.

I think lighting would be a good one to start with. Some questions:
- How will users save/load lighting settings?
- Should the user be able to lock the sunlight to the model so that the lighting conditions remain the same as the model moves around the virtual lab?
[...]

PR #1234, opened 2020-10-29
Scripting shell cmds to visualizer

aymanhab:
Fixes issue #0

### Brief summary of changes
Support sending options from scripting shell of the GUI to the visualizer (so far supports changing video format to save as tar file or animated gif (png, jpg,gif default webm) or override frame rate (default 30)
### Testing I've completed
try either of these in scripting shell:
setVisualizerOption("video_format", "png") // supported gif, jpg, png
setVisualizerOption("frame_rate", "60")
### CHANGELOG.md (choose one)

- updated...


aymanhab:
Build ready to test @carmichaelong @jenhicks @nickbianco

aymanhab:
@tkuchida You maybe interested in this as well. We can easily leverage the same mechanism to set other options (e.g. Camera position, or other customizations as you were doing for the book, what would be the top custom settings that you ended up using? Thank you

tkuchida:
> what would be the top custom settings that you ended up using?

@aymanhab #1137

carmichaelong:
@aymanhab some notes below:

Tested on Win 10, using Gait2392 and a crouch motion from Tutorial 1

1. Testing jpg, png, and gif all worked.
2. Cool that you get a moving .gif automatically out at the end!
3. Testing different frame rates with jpg didn't seem to change the output (the number of frames output were roughly the same)
4. Most frames were fully drawn, but did see at least one frame that was not:
image
5. Un-tar-ing is easy enough for me. I wonder if you could just dump the files in a folder as specified by the user (to save the extra step).

aymanhab:
Any update on this PR @carmichaelong ? Do you want to give it a shot @tkuchida Will need to update changelog once we agree on what to expose.

carmichaelong:
@aymanhab Some notes from further testing:
- The "frame_rate" option seems to have an effect for "webm" and "gif" options, but the frames seem to be played too quickly on the .gif version (when set to 60 or 120)
- The "frame_rate" option did not have any noticeable effect on .jpg or .png options (i.e., the number of frames in the .tar file were roughly the same). I even tried with a frame_rate as low as 3 or as high as 120.
- On a slower machine, I tried loading 3 models and motions and recording. All frames did have a full picture (unlike above), however, I haven't tried to stitch the frames together in a video to see if there is any "hitching" apparent. I imagine hitching can happen still even if the frame rate is still correct.

tkuchida:
@aymanhab What did you need me to test?

aymanhab:
Thanks for the feedback @carmichaelong

aymanhab:
@tkuchida I mainly wanted to see if you can exercise these new options and if you'd find the different video formats or frame rates useful since you're our most experienced user of the visualiser, but nothing urgent. I looked into your list and indeed most large scale changes require recreating the scene so can't be done on the fly.

tkuchida:
> I looked into your list and indeed most large scale changes require recreating the scene so can't be done on the fly.

Perhaps a way around this could be to set new properties and then destroy/respawn just the visualizer window? That would at least be better than restarting the whole GUI. Another possibility could be to create a standalone "render mode" window for generating pretty images with adjustable lighting, etc. Not sure where any of this lands on your roadmap but unfortunately it isn't on mine at the moment.

carmichaelong:
@aymanhab Tested stitching some of the frames together for a video. I took the base ToyDropLanding model to create a 5 second forward simulation. I then set the output for "png". The output frames generated 73 frames (which is ~15 fps). Using `setVisualizerOption("frame_rate", "30")` also had roughly the same number of frames (74 frames, which may depend on how long before I stop recording after the motion stops).

Stitching was done with ffmpeg: `ffmpeg.exe -framerate 30 -i %07d.png -vf scale=-2:404 pix_fmt yuv420p output.mp4`

Frames and video: [toydroplanding.zip](https://github.com/opensim-org/opensim-gui/files/5591949/toydroplanding.zip)

aymanhab:
Thanks @carmichaelong while I don't know what the ffmpeg options exactly do, my conclusion is that it's still useful to generate frames so that you can pull a single snapshot and make an mp4 (if using ffmpeg or other third-party tools). I didn't document frame_rate option in the changelog because I think it's less useful based on your testing. 

With that and considering that we move in the right direction I'd lean to merge this PR unless you see a reason not to.

carmichaelong:
LGTM to merge this. The ffmpeg options were more for me to document what I had done to stitch things together so that I don't have to look certain options up again (sorry if that was confusing!). That being said, maybe I should have concluded with the following: definitely useful to output frames, but it seems like the frames are pretty sparse (at least on a 7 year old laptop) even before going into ffmpeg, which can be addressed in a later PR.

carmichaelong (approved):

aymanhab:
Thanks @carmichaelong much appreciated, will merge. 

One follow up note is that by changing the playback speed in the GUI we can have more poses that are closer (in simulation time) resulting in more dense images, then ffmpeg can fix the time stretch. For this scenario there would be no need to change frame_rate in the scripting shell altogether, I think.

carmichaelong:
> changing the playback speed in the GUI we can have more poses that are closer (in simulation time) resulting in more dense images, then ffmpeg can fix the time stretch

I just tried that and it does work, though there' something weird about the number of frames (i.e., it doesn't linearly scale). At 1.0x playback, I got ~75 frames (or ~15 fps), and then at 0.2x playback I got ~300 frames (or ~60 fps), so 4x the frames rather than 5x.

PR #1257, opened 2021-01-09
Make toolbar background opaque for osx catalina

aymanhab:
Fixes issue #0

### Brief summary of changes
Toolbar for motion control appears black by default on os-x Catalina, attempt fix

### Testing I've completed
Ran locally
### CHANGELOG.md (choose one)

- updated...


aymanhab:
While not completely back to pre-Catalina, the toolbars are usable, will continue investigating why the background border turned dark, otherwise we can start testing..

Issue #1260, opened 2021-01-21
Plotter dialog icon is missing in 4.2 Beta


aymanhab:
### Steps to reproduce
Open plotter window on Windows machine, the window should have OpenSim logo, it shows with Java/Duke icon instead

### Actual result
![CapturePlotter42](https://user-images.githubusercontent.com/3972208/105420373-1c9b7f80-5bf5-11eb-9649-1b6c50607da0.JPG)

(include screen grabs where appropriate)

PR #1264, opened 2021-01-28
Replace reference to old/missing image for dialog icon

aymanhab:
Fixes issue #1260

### Brief summary of changes
Dialogs were not finding icon file and ending with generic java icon, fixed to find correct icon on path.
### Testing I've completed
Built and launched dialogs(Plotter, Excitation Editor) both now show correct icon
### CHANGELOG.md (choose one)

- no need to update because minor bugfix


aymanhab (commented):
Trivial code change to find correct icon file

Issue #1268, opened 2021-02-09
Random exception while editing scale tool setup 


aymanhab:
Reported by user testing and by  Antoine during testing 4.2 Beta

### Actual result

(include screen grabs where appropriate)

aymanhab:
Should be fixed by PR #1268

aymanhab:
Waiting for verification!

jenhicks:
@aymanhab I could test this, but can't find specific details about what to test.

aymanhab:
Tested by Antoine, and he gave the thumbs up, so no need to worry about it.

PR #1269, opened 2021-02-11
Update bindings and fix GUI to make standalone copies of tasks

aymanhab:
Fixes issue #1268 

### Brief summary of changes
Make local copies of IKTask(s) for GUI use so they don't get garbage collected using copy constructors (instead of using clone)

### Testing I've completed
Built GUI and couldn't cause the issue reported by Antoine to reproduce.

### CHANGELOG.md (choose one)

- no need to update because internal bugfix


aymanhab:
@antoinefalisse if you can test the artifact for a fix to the scaletool exception that would be great. Thank you

Issue #1272, opened 2021-02-11
Unable to avoid heading correction in the GUI


aymanhab:
In IMU Placer, unable to omit heading correction unless a saved setup file omitting the correction is loaded (can't simply deselect name/axis from default IMU Placer Tool settings or from a loaded setup file with a correction). Error: Invalid specification of heading aixs ' ' found.

### Actual result

(include screen grabs where appropriate)

aymanhab:
The error message was due to passing space ' '  instead of empty string '' to the tool.

jenhicks:
So can this be closed?

aymanhab:
If we merge the branch and the build on master works as expected (which should be the case) then yes.

Issue #1273, opened 2021-02-11
IMU IK weight issues


aymanhab:
- Should IMU IK weights be saved in the settings file? If so, not currently happening
- After running the IK tool, Sensor Weights window is blank when opened. Opening Sensor Weights window multiple times before running the tool works fine.
- No evidence that manually changing weights affects output - hip, knee, ankle angles stayed the same on a couple different weight set runs
- ### Steps to reproduce

### Actual result

(include screen grabs where appropriate)

PR #1275, opened 2021-02-17
Fix imu gui issues 0221

aymanhab:
Fixes issue #1272, #1273

### Brief summary of changes
Create a  copy constructor for OrientationWeightSet and utilize it to keep GUI side copy of OrientationWeightSet for editing and setting back in tool.

### Testing I've completed
Scenarios described in issues 1272, 1273 work as expected now.

### CHANGELOG.md (choose one)

### DIFF
--- a/Gui/opensim/modeling/src/org/opensim/modeling/OrientationWeightSet.java
+++ b/Gui/opensim/modeling/src/org/opensim/modeling/OrientationWeightSet.java
@@ -58,7 +58,11 @@ public class OrientationWeightSet extends SetOientationWeights {
   }
 
   public OrientationWeightSet() {
-    this(opensimActuatorsAnalysesToolsJNI.new_OrientationWeightSet(), true);
+    this(opensimActuatorsAnalysesToolsJNI.new_OrientationWeightSet__SWIG_0(), true);
+  }
+
+  public OrientationWeightSet(OrientationWeightSet arg0) {
+    this(opensimActuatorsAnalysesToolsJNI.new_OrientationWeightSet__SWIG_1(OrientationWeightSet.getCPtr(arg0), arg0), true);
   }
 
 }
--- a/Gui/opensim/modeling/src/org/opensim/modeling/opensimActuatorsAnalysesToolsJNI.java
+++ b/Gui/opensim/modeling/src/org/opensim/modeling/opensimActuatorsAnalysesToolsJNI.java
@@ -2163,7 +2163,8 @@ public class opensimActuatorsAnalysesToolsJNI {
   public final static native String OrientationWeightSet_getClassName();
   public final static native long OrientationWeightSet_clone(long jarg1, OrientationWeightSet jarg1_);
   public final static native String OrientationWeightSet_getConcreteClassName(long jarg1, OrientationWeightSet jarg1_);
-  public final static native long new_OrientationWeightSet();
+  public final static native long new_OrientationWeightSet__SWIG_0();
+  public final static native long new_OrientationWeightSet__SWIG_1(long jarg1, OrientationWeightSet jarg1_);
   public final static native void delete_OrientationWeightSet(long jarg1);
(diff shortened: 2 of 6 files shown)


aymanhab:
This PR will not build on ci until core PR https://github.com/opensim-org/opensim-core/pull/2955 is merged.

aymanhab:
@carmichaelong you can test/verify fixes to issues #1272, #1273 in the artifacts of this PR when done.

carmichaelong:
Re #1272:
Both models started with loading the original IMU placer setup file. Left model has the heading correction "disabled", right model uses the setup file as is.
avoid_heading_correction

Re #1273:
I'm having trouble verifying because it keeps crashing when I try to run IMU IK with the provided files (and after running IMU Placer)

aymanhab:
Thanks for testing @carmichaelong 
1. It appears heading correction indeed can be disabled or bypassed so we can put this one to rest.
2. Can you specify weights and can they be saved to a file?
3. What files are you using that produce a crash?

aymanhab:
Also can you indicate which build/artifact you're using since there was a bugfix submitted on master that could be related to random crashes ☹️

carmichaelong:
> Can you specify weights and can they be saved to a file?

Yes, I can adjust the weight, save it to a setup file, and then loading the setup file has the right weights in the GUI. However, one thing I saw was that the weights only accepted numbers between 0 and 1 (while for the usual IK they can be any positive number). Of course users can scale their problem to have the same effect, but might be good to be consistent with IK.

> What files are you using that produce a crash?

I'm using the files distributed with the GUI (Rajagopal_OpenSense)

carmichaelong:
I used the artifact from this PR (OpenSim 4.2-2021-02-17-a00ef82)

aymanhab:
@carmichaelong Thanks I was able to reproduce the crash in my dev. environment, so I'm investigating now.
The fact that weights are 0-1 is baked into the Tool so we can't change though I agree it's a bit inconsistent with other IK. Unless we're mixing both Marker and IMU based IKs only the relative weight matters so shouldn't be an issue for users AFAICT.

carmichaelong:
> The fact that weights are 0-1 is baked into the Tool so we can't change though I agree it's a bit inconsistent with other IK. Unless we're mixing both Marker and IMU based IKs only the relative weight matters so shouldn't be an issue for users AFAICT.

Sounds good, and good point about if we mix the weights down the road. The nice thing long term is that the current implementation is stricter than the alternative, so if we ever go to the alternative, it shouldn't break current files.

aymanhab:
The crash has been fixed so you can try again @carmichaelong Thanks for testing.

carmichaelong:
Thanks @aymanhab. A few more comments:

> From #1273: "After running the IK tool, Sensor Weights window is blank when opened. Opening Sensor Weights window multiple times before running the tool works fine."
I saw this happen one time, but I can't replicate this reliably.

I'm also having trouble running a new IMU IK setup file with non-default weights. I made the attached setup file by using the Rajagopal_OpenSense example, and then changing one of the weights. When I load this file, the Run button is grayed out.
[test.zip](https://github.com/opensim-org/opensim-gui/files/6007179/test.zip)

I have also seen some random crashes while running IMU IK with non-default weights, however, I'm also having issues with reproducing this reliably.

aymanhab:
@carmichaelong I fixed the validation to be invoked on load (works already if you browse for the sensor orientation file), which makes the run button enabled on loading your test.xml, however I can't reproduce any crashes. Is now a reasonable point to pass the build to others to try to reliably reproduce crashes (which may or maynot be related to this change)? Thank you.

carmichaelong:
Sounds good, yes I think now could be a reasonable point to see if others have issues too.

aymanhab:
@jenhicks @nickbianco and @antoinefalisse  you can use the artifact from this PR to test the latest changes, it also includes the copy bug-fix that was merged lately into opensim-core so keep an eye on random crashes when editing or using the tools. Thank you

aymanhab:
Merging based on earlier comment from @carmichaelong to make available for further testing.

Issue #1274, opened 2021-02-16
Specifying default states file in Forwardtool affects end time


aymanhab:
### Steps to reproduce
Reported by Carmichael in testing party
### Actual result

(include screen grabs where appropriate)

PR #1277, opened 2021-02-23
Leave time range alone when loading initial states file.

aymanhab:
Fixes issue #1274

### Brief summary of changes
Leave tool time alone when loading initial states file, the tool takes care of adjusting for initial states time downstream.

### Testing I've completed
Built GUI, tested that forward tool time is not affected when default states file is selected.

### CHANGELOG.md (choose one)

- no need to update because minor bugfix


aymanhab:
This is a rather trivial change that disable time change in one scenario, we can test post merge.

PR #1298, opened 2021-04-19
Minor linux GUI layout improvements

halleysfifthinc:
### Brief summary of changes

There are 2 layout problems in the Linux build (at least with Gnome/GTK) when compared to Windows:
1. All sliders show their current value above the slider, which uses more vertical space and pushes the slider/track down, cutting off the top and bottom of the slider value and slider track, respectively. (See first image)
2. Text size in text fields is too large to show all the text, which leads to cut off text which is hard to read. (Check out the coordinate value text field and the "Time" and "Speed" text in the motion control panel.)

Additionally, I added a tool tip for the current motion text field, which also cuts off filenames which are too long, and I changed the font for the "Simulate" and "Motion" labels to be based on the system font but bolded. The previous font was small in comparison to the rest of the text in the GUI.

#### Before
image
image

#### After
image
image

### Testing I've completed

I don't have a build environment setup for Windows, so I will use the Windows build generated by CI to look for undesired changes on Windows; I don't have access to a Mac.

### CHANGELOG.md

Not sure if/what y'all would want in the changelog, so I will leave that alone for now.

### DIFF
--- a/Gui/opensim/coordinateViewer/src/org/opensim/coordinateviewer/CoordinateSliderWithBox.form
+++ b/Gui/opensim/coordinateViewer/src/org/opensim/coordinateviewer/CoordinateSliderWithBox.form
@@ -21,9 +21,9 @@
       <Group type="103" groupAlignment="0" attributes="0">
           <Group type="102" alignment="0" attributes="0">
               <Component id="jCoordinateNameLabel" min="-2" pref="100" max="-2" attributes="0"/>
-              <EmptySpace min="-2" max="-2" attributes="0"/>
-              <Component id="jFormattedTextField" min="-2" pref="55" max="-2" attributes="0"/>
-              <EmptySpace min="-2" pref="0" max="-2" attributes="0"/>
+              <EmptySpace max="-2" attributes="0"/>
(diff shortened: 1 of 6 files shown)


halleysfifthinc:
Looks like this makes the Windows text fields rather large:
image

I will see if I can figure out a solution that doesn't mess with the Windows layout as dramatically.

aymanhab:
Thanks for working on this @halleysfifthinc 👍 
We're getting the next beta release assembled and I'm wondering if it makes sense to include this PR considering we have only few users of the GUI on linux, and it has the potential to change the layout on Windows and OSX. Please let me know your thoughts.

halleysfifthinc:
I don't think this ready for merging yet, given the unwanted/disproportionate changes in the layout on Windows. However, I will try to finish this sometime soon.

halleysfifthinc:
Text field width in the motion control panel and coordinate viewer is now set to "4", which confusingly shows 5 digits comfortably.
I also changed the alignment of the current/start/end time text fields to right-aligned.

### Motion control panel
image
image

### Coordinate viewer
image
image

aymanhab:
Merging to a branch that we can build/test as opposed to fork so we say the effect on all platforms.

PR #1317, opened 2021-09-29
use updated branches in threejs and opensim-visualizer repos

aymanhab:
Fixes issue #0

### Brief summary of changes
Upgrade three.js to use the latest version 126
### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/Gui/opensim/build.xml
+++ b/Gui/opensim/build.xml
@@ -168,6 +168,9 @@
         <copy todir="${opensim.threejs.dir}/examples/js">
             <fileset dir="threejs/examples/js" />
         </copy>
+        <copy todir="${opensim.threejs.dir}/examples/jsm">
+            <fileset dir="threejs/examples/jsm" />
+        </copy>
     </target>
     <target name="build-mac-installer" depends="build-mac,copy-visualizer,copy-models">
         <property name="lib.dir" value="/Users/aymanhabib/dev/install/opensim-core-install/lib"/>


aymanhab:
Items to test:
[ ] launch application on Windows, load pendulum model
[ ] Repeat for osx
[ ] Close model
[ ] Exercise toolbar view buttons (views +x,-x,+y,-y,+z,-z), zoom-in/out/refit
[ ] Exercise record snapshot, movies
[ ] Exercise visualizer controls menu (background options, floor options, lighting options)
[ ] Exercise model context menus: offset, COM
[ ] Exercise object/geometry menus (color, wireframe, transparency, show_axis) with undo
[ ] Selection in tree and visualizer are consistent
[ ] Visual editing of Markers, PathPoints
[ ] Load motion, play, associate Marker Data, Force Data, Sensor Data
[ ] Synchronize motions
[ ] Load/close experimental and sensor data, change size

aymanhab:
@jenhicks Once this build is successful we can look at testing to account for what I've done myself as part of testing this PR, but this is the list of items to test.

aymanhab:
Build is ready if anybody can give it a spin @jenhicks  @carmichaelong @nickbianco

aymanhab:
@carmichaelong The "dots" were handles to move the model light, I removed them. Thanks for the find.
The undo for ShowAxis and COM display are bugs unrelated to the threejs upgrade (reproduce in 4.3) will open issues to prioritize relative to other issues but the change will be in opensim-gui code base not the visualizer.
I meant to test model-> Display ->Model offset... functionality
For the last comment to change size, you can modify the display size of Experimental Markers and Force Arrows

aymanhab:
Testing on osx shows no ill-effects at all, everything worked as expected. The main issue I'd wait for is tuning the washed out colors we observed that could be due to change in lighting.

carmichaelong:
@aymanhab Thanks, the colors look better now than before. Will be good to get other feedback if further tweaks are necessary (it still looks different than previous versions which might be OK. Other issue is that I'm realizing is that I'm looking at this on a different monitor now too).

The force color and size updates are now working with both cases of 1) Preview experimental data and 2) Associate motion data (with a current motion). One small bug when you use the "Associate motion data route":
1) Load gait10dof18
2) Load CMC results
3) "Associate motion data" with the GRF file in the base folders.
4) The properties pane shows the "Force display size" as 0.001, although it should say 1.0 (since it seems like setting the size to 1.0 doesn't change arrow size, and setting smaller than 1.0 leads to smaller arrows)

aymanhab:
@carmichaelong Can you check that all issues coming out of this are opened as a separate issues so that we can close this PR and reduce the clutter? Thank you

PR #1332, opened 2022-03-04
Update core bindings

aymanhab:
Fixes issue #0

### Brief summary of changes
Rerun swig, update bindings

### CHANGELOG.md (choose one)

- updated...


aymanhab:
This is updating autogenerated files, plus work to fix ci which is justified by the build success vs master failing.

PR #1333, opened 2022-03-18
Update swig generated java files

aymanhab:
Fixes issue #0

### CHANGELOG.md (choose one)


aymanhab:
folded into another PR. closing

PR #1328, opened 2021-12-03
Fix undo visuals1319

aymanhab:
Fixes issue #1319 
### Brief summary of changes
Add blocks to support undo/redo of "show COM", "show Axes"

### Testing I've completed
Visual inspection/test of behavior.

### CHANGELOG.md (choose one)
- updated...

PR #1336, opened 2022-03-28
Minor linux GUI layout improvements (#1298)

aymanhab:
* Fix undo visuals1319 (#1328)

* Fix undo/redo of com visualization

* Support undo/redo of FrameToggleVisibilityAction

* Update CHANGELOG.md

Description of changes made in this PR and others earlier in 4.4 stream.

* include visualizer repo update

* Update continuous-integration.yml

No caching of builds for now due to osx upgrade

* Fix coordinate value text being cutoff by too short text field

* Fix text overflow in short TextFields; Adjust font & tooltip for motions

- Start, current, end time and speed text fields were too short to
accomodate 3 digits of seconds with 3 decimal precision
- "Motion" text was very small, changed to derive from default system
font but bolded
- Motion name label overflows for long file names; add dynamic tooltip
to show complete filename

* Update 'Simulate' text label to derive from system font but bolded

* Fix look & feel bug with Gnome (GTK) where slider values are always shown

* Fix (properly) text field width by assigning preferred num columns

* Apply similar fix to coordinate viewer

Co-authored-by: Ayman Habib <ahabib@stanford.edu>

Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated...


aymanhab:
Port/test fixes from @halleysfifthinc

aymanhab:
@halleysfifthinc @carmichaelong Can we check the Look and Feel on OSX and linux of the artifact from this PR before merging? Looking good on windows but I'm worried about osx inadvertently affected. Is there more to testing than checking the look-and-feel and behavior of motion slider and coordinate sliders?

halleysfifthinc:
I will take a look at the Linux artifact tonight. Only the look/layout is changed, so there shouldn't be (couldn't be, to the best of my knowledge) any changes to slider behavior.

carmichaelong:
On OSX, tried loading some models and motions and moving sliders around. All of that seemed to work well! Didn't notice much difference between 4.3 and this version, however, one big plus is that with my computer on dark mode, the font colors chosen are much better (seems like it chooses a typical light mode set instead of attempting to do a half-filled-dark-mode)!

Noticed a couple of issues, but all of these existed in 4.3 (i.e., 1) if the window width is too narrow, the panels in the top bar are pushed to two rows and can't be accessed well, 2) in the Coordinates tab, "Values" and "Speed" are not centered over the corresponding boxes).

halleysfifthinc:
Everything still looks good. (Pics in #1298 are still accurate).

(Though I will note that the Linux artifact is broken due to the adolc libs not being copied correctly [opensim-core#2930](https://github.com/opensim-org/opensim-core/pull/2930))

aymanhab:
Thanks @halleysfifthinc will merge, if you can assess the fix for adolc or suggest a best fix, that would be awesome. 👍

Issue #1340, opened 2022-04-12
Updating IMU's Translation and Rotation properties don't update until model is saved and loaded again


carmichaelong:
@aymanhab 

### Steps to reproduce
1. Run IMU placer step on OpenSense example
2. Change properties for Translation and Rotation for one of the IMUs (e.g., `femur_r_imu`)
3. Note that the GUI does not update
4. Save the model
5. Load the model
6. GUI shows updated IMU location

Checking some other components, did not find other components that had the same behavior. e.g., muscle path updates, wrapping surface location, and even updating PhysicalOffsetFrame's within a Joint showed the current update immediately in the GUI

### Environment and GUI version
macOS 12.3.1. OpenSim 4.4-2022-04-11-e3e3d1d

aymanhab:
Thanks for reporting @carmichaelong This boils down to the fact that the IMU is "connected" to the Frame rather than a child of it. The axes for example update instantly as expected. Can you verify please?
Will investigate how to collect components that are connected to a Frame or some other surgical workaround.

carmichaelong:
Yes, confirmed that the axes update instantly.

PR #1341, opened 2022-04-14
Full refresh of model decorations on Frame edit 

aymanhab:
Make sure to update components that are "connected-to" rather than only those "owned" by a Frame

Fixes issue #1340

### Brief summary of changes
When a Frame location/orientation are edited, refresh all model decorations rather than only subcomponents or owned components since there are other components that are connected to Frames via sockets (e.g. IMU)
  
### Testing I've completed
Tested model/change in issue #1340 

### CHANGELOG.md (choose one)
- updated...


aymanhab:
@carmichaelong ready to test

aymanhab:
@jenhicks or @nickbianco if you can test it would be great since @carmichaelong is out. Thank you

carmichaelong:
@aymanhab LGTM. Tested the Windows 10 artifact. IMU visualization changed for updates to both Translation and Rotation

aymanhab:
Thanks @carmichaelong 👍

Issue #1344, opened 2022-04-19
Empty file after saving experimental TRC file


mrrezaie:
### Steps to reproduce
1. in GUI, go to File > Preview Experimental Data.
2. Select subject01_walk1.trc from Gait2392_Simbody sample data.
3. Unfold it in the Navigator tab.
4. Select Save As and save the data as a new TRC file.
5. (Also if save it through Transform)
### Actual result
The saved file is empty and there is  an Exception:
`java.lang.IllegalArgumentException: No enum class org.opensim.modeling.Units$UnitType with value 771
	at org.opensim.modeling.Units$UnitType.swigToEnum(Units.java:106)
	at org.opensim.modeling.Units.getType(Units.java:70)
	at org.opensim.view.experimentaldata.AnnotatedMotion.saveAsTRC(AnnotatedMotion.java:428)
	at org.opensim.view.experimentaldata.AnnotatedMotion.saveAs(AnnotatedMotion.java:393)
	at org.opensim.view.experimentaldata.ClassifyDataJPanel.saveTransformedMotion(ClassifyDataJPanel.java:271)
	at org.opensim.view.motions.MotionsSaveAsAction.saveMotion(MotionsSaveAsAction.java:45)
	at org.opensim.view.motions.MotionsSaveAsAction.performAction(MotionsSaveAsAction.java:80)
	at org.openide.util.actions.CallableSystemAction$1.run(CallableSystemAction.java:105)
	at org.openide.util.actions.ActionInvoker$1.run(ActionInvoker.java:70)
	at org.openide.util.actions.ActionInvoker.doPerformAction(ActionInvoker.java:91)
	at org.openide.util.actions.ActionInvoker.invokeAction(ActionInvoker.java:74)
	at org.openide.util.actions.CallableSystemAction.actionPerformed(CallableSystemAction.java:102)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.AbstractButton.doClick(AbstractButton.java:376)
	at javax.swing.plaf.basic.BasicMenuItemUI.doClick(BasicMenuItemUI.java:842)
[...]

PR #1346, opened 2022-04-21
Prserve rates as ints rather than doubles to make header parsing robust

aymanhab:
Fixes issue #1344

### Brief summary of changes
Write frameRate, cameraRate etc. as ints to avoid confusing the parser built into MarkerData class (GUI doesn't have its own parser). Ideally we switch to the TimeSeriesTable to avoid duplicity but those need work to be usable by the GUI.

### Testing I've completed
Tested workflow suggested in issue and it all worked.

### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/experimentaldata/AnnotatedMotion.java
+++ b/Gui/opensim/view/src/org/opensim/view/experimentaldata/AnnotatedMotion.java
@@ -72,9 +72,9 @@ public class AnnotatedMotion extends Storage {
     private double unitConversion = 1.0;
     private boolean boundingBoxComputed=false;
     private double[] currentRotations = new double[]{0., 0., 0.};
-    private double frameRate=0.;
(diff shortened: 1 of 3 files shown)


nickbianco:
Is it possible that any of these values can take non-integer values? Or do they get converted to ints when passed to the MarkerData class?

Otherwise, LGTM.

aymanhab:
@nickbianco In practice I've never seen rates that are non-ints, the MarkerData class uses doubles internally but reads/prints using custom simmIO class that's likely not robust and the long term plan should be to retire it altogether.

nickbianco:
Okay, sounds good. Seems like this a find stop-gap solution until replace the simmIO class later.

aymanhab:
> Okay, sounds good. Seems like this a find stop-gap solution until replace the simmIO class later.

Agreed, will wait on more testing before merging. Thanks again @nickbianco

Issue #1168, opened 2020-01-19
Previewed experimental markers disappear when model pose is changed


tkuchida:
### Steps to reproduce
- open arm26.osim
- File -> Preview Experimental Data…; select `Resources\Models\Arm26\InverseKinematics\arm26_elbow_flex.trc`
- in Navigator panel, double-click "arm26" to re-select model
- in Coordinates panel, click Poses -> Default

### Expected result
Nothing should change in the visualizer.

### Actual result
Experimental markers disappear.

### Environment and GUI version
Windows 10, OpenSim 4.0 public release

jenhicks:
@aymanhab will take a look.

aymanhab:
Proposed fix in PR

PR #1347, opened 2022-04-22
Leave experimental data alone when cleaning up motion objects 

aymanhab:
Leave experimental data alone upon motion change

Fixes issue #1168

### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/Gui/opensim/view/src/org/opensim/view/motions/MasterMotionModel.java
+++ b/Gui/opensim/view/src/org/opensim/view/motions/MasterMotionModel.java
@@ -33,6 +33,7 @@ import javax.swing.event.ChangeListener;
 import javax.swing.event.EventListenerList;
 import org.opensim.modeling.Model;
 import org.opensim.modeling.Storage;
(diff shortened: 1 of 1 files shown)


aymanhab:
@tkuchida Would you have a chance to check this has been fixed in this PR artifact? Thank you

tkuchida:
The Issue that was raised in #1168 has been fixed. Minor note: The muscles change color after clicking Poses -> Default (see below). Not sure why… presumably the location of a light is being updated, but it seems like this type of update should occur when the model gets the focus back (i.e. after double-clicking "arm26" in the Navigator) rather than when Poses -> Default is clicked.

Before clicking Poses -> Default | After clicking Poses -> Default
--- | ---
image-before | image-after

aymanhab:
Muscle changing color when state is changed since 4.0 e.g. by manually changing coordinates is expected behavior, otherwise you have inconsistent muscle colors and coordinates.

aymanhab:
Thanks for testing @tkuchida

Issue #1351, opened 2022-05-16
wrist.osim model does not open


carmichaelong:
@aymanhab 

Model wrist.osim provided with installer opens in 4.3 without any problems. Does not open with the 4.4 beta GUI.

```
java.lang.IllegalArgumentException: Illegal Capacity: -1
	at java.util.ArrayList.<init>(ArrayList.java:157)
	at org.opensim.threejs.ModelVisualizationJson.createJsonForGeometryPath(ModelVisualizationJson.java:1704)
	at org.opensim.threejs.ModelVisualizationJson.processDecorationsForComponent(ModelVisualizationJson.java:509)
	at org.opensim.threejs.ModelVisualizationJson.createJsonForModel(ModelVisualizationJson.java:467)
	at org.opensim.threejs.ModelVisualizationJson.<init>(ModelVisualizationJson.java:427)
	at org.opensim.view.pub.ViewDB.getJsonForModel(ViewDB.java:1147)
	at org.opensim.view.pub.ViewDB.update(ViewDB.java:401)
	at java.util.Observable.notifyObservers(Observable.java:159)
	at org.opensim.view.pub.OpenSimDB.addModel(OpenSimDB.java:153)
	at org.opensim.view.pub.OpenSimDB.addModel(OpenSimDB.java:141)
	at org.opensim.view.FileOpenOsimModelAction$2.run(FileOpenOsimModelAction.java:100)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:728)
	at org.netbeans.core.TimableEventQueue.dispatchEvent(TimableEventQueue.java:136)
[catch] at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
[...]

PR #1352, opened 2022-05-17
Avoid special handling of path-segments with conditional path points

aymanhab:
Fixes issue #1351

### Brief summary of changes
Fix indexing issue when passing from one path-segment to the next (restore default behavior)
### Testing I've completed
Tested other models with visualization issues with conditional path points and multiple wrapping
### CHANGELOG.md (choose one)

- no need to update because this was a side-effect of earlier fix/internal

### DIFF
--- a/Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
+++ b/Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
@@ -1688,13 +1688,6 @@ public class ModelVisualizationJson extends JSONObject {
                 mapUUIDToComponent.put(pathpoint_uuid, cpp);                
                 pathpointActive_jsonArr.add(false);
                 pathpoint_jsonArr.add(pathpoint_uuid.toString());
-                // traverse backward to an active point
-                /*
-                int tempIndex = ppointSetIndex;
-                while (!pathPointSetFromProperty.get(tempIndex).isActive(state))
-                    tempIndex--;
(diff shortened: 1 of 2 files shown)


Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -1689,11 +1689,12 @@ else if (secondIndex == -1){ // Conditional Path point that's inactive
                 pathpointActive_jsonArr.add(false);
                 pathpoint_jsonArr.add(pathpoint_uuid.toString());
                 // traverse backward to an active point
+                /*
```
carmichaelong, line 1692: If this is no longer needed, would be better just to delete this section. If this is helpful to keep around, it would be good to add a comment about what this code was attempting to solve and why this didn't work.

carmichaelong (changes requested):
Testing on Mac showed that this did indeed fix the issue with loading the wrist model. I left one comment to make sure that the code is either cleaned up or has appropriate comments for maintainability later.

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
aymanhab: Deleted, thanks @carmichaelong

carmichaelong (approved):

aymanhab:
Thanks @carmichaelong 👍

PR #1353, opened 2022-05-20
Welcome page, Doxygen link and splash screen

aymanhab:
Fixes issue #0

### CHANGELOG.md (choose one)

- updated...


carmichaelong:
Thanks @aymanhab, looks good. Just one question on the doxygen link since it doesn't exist yet. What's the normal order for changing that link and updating doxygen? Might be confusing for some users testing the beta if the doxygen link doesn't exist before it gets released.

aymanhab:
Good point, we ask Henry to upload/update separately. I'll send him the zip
file now but we need the most recent since the new Moco classes need to be
there.

aymanhab:
It exists now thanks to Henry, let me know if you approve merging. Thank you

carmichaelong (approved):

aymanhab:
Thanks @carmichaelong

Issue #1354, opened 2022-05-23
IMU InverseKinematics GUI throws exception if weights are not visited


aymanhab:
### Steps to reproduce
Follow opensense demo and try to run IK without visiting/clicking the weights button.
### Expected result
IK runs
### Actual result
Exception with no message dialog pops up, reported by Nev.Pires from Vicon
![image](https://user-images.githubusercontent.com/3972208/169869370-79e13034-5e93-4b34-a039-d8a3cc143d96.png)

(include screen grabs where appropriate)

carmichaelong:
@aymanhab which demo / files were you using? using the Rajagopal_OpenSense setup files does not reproduce this issue for me

AlbertoCasasOrtiz:
@carmichaelong I also got this error while testing this tutorial. To reproduce it:

1. Load the model "Rajagopal2015_opensense.osim".
2. Go to "Tools -> IMU Inverse Kinematics..."
3. Select "walking_orientations.sto" as Sensor Orientation File.
4. Click Run.

I am using the following build: OpenSim 4.4-2022-05-17-cc0a2cf

carmichaelong:
Thanks @AlbertoCasasOrtiz, that was helpful! Just for documenting if we ever look back at this issue: the model needs to have IMUs on it (e.g., a model from after running IMU Placer), and this workflow likely is different than testing before since this does not simply load the given setup file.

aymanhab:
@carmichaelong Yes the difference is here we run the tool live from scratch without checking the weights (which is probably what we all did when testing) instead of loading  a file.

PR #1355, opened 2022-05-23
Fix issue 1354 exception if no weights button before IK

aymanhab:
Fixes issue #1354 
### Brief summary of changes
Weights were initialized to null instead of using the getter that populates them if null. 
### Testing I've completed
Ran workflow in issue without exception being thrown
### CHANGELOG.md (choose one)

- not sure if it is worth including in change log considering it's a rare scenario


carmichaelong (approved):

carmichaelong:
Approved based on reproducing error in the linked issue on an earlier build, and that the error does not occur in this PR's build

aymanhab:
Thanks @carmichaelong 👍

Issue #1373, opened 2022-06-28
OpenSim not appearing in Ubuntu applications menu


AlbertoCasasOrtiz:
### Steps to reproduce

Install the application and try to start it from the applications menu.

### Expected result

The OpenSim icon appears in the applications menu and you can start the application from there.

### Actual result

OpenSim is not appearing in the applications menu. However, if you check the path `/usr/share/applications`, the file `opensim.desktop` is there.

### Environment and GUI version

- GUI version: Built from Master and last artifact for Linux([1206](https://github.com/opensim-org/opensim-gui/actions/runs/2566453821))
- Environment: Ubuntu 22.04 and Ubuntu 18.04.

AlbertoCasasOrtiz:
Hello @rutuja1409 

Looks like you have found two problems with the INSTALL script.

The first one (the condition with 18.04 hardcoded) is already fixed and waiting for merge in #1391. As you discovered, a workaround is to change it to 20.04.

The second one is caused by a bug in the script. When the INSTALL script executes the chown command in the [following line](https://github.com/opensim-org/opensim-gui/blob/47f8d014edabf0f08bf31e4d7e2c20c2c7bb63ae/Gui/opensim/installer_files/Linux/INSTALL#L112), it needs the name of the user and the name of the group. However, looks like we are not retrieving the name of the group to which the user belongs, but instead we are using the name of the user as the name of the group.

 In your case, looks like the name of the group is different from your user. 
 
 You can get the name of the group your user belongs with the following command:
` id -nG <your-user>`

You will get as output the list of groups to which your user belongs. Your main group should be the first one.
 
To solve your problem, try changing the following line:

https://github.com/opensim-org/opensim-gui/blob/47f8d014edabf0f08bf31e4d7e2c20c2c7bb63ae/Gui/opensim/installer_files/Linux/INSTALL#L112

Substituting it by the next command should work for you:

`sudo chown -R <your-user>:<your-main-group> $installPrefix`

Where `<your-user>` should be **lahoti** and `<your-main-group>` should be the first one appearning after executing ` id -nG <your-user>`.

Please, let me know if this works.

AlbertoCasasOrtiz:
Closed with #1392

PR #1374, opened 2022-06-28
Fix installer ubuntu

AlbertoCasasOrtiz:
Fixes issue #1373

### Brief summary of changes

Changed INSTALL script so the `opensim.desktop` file is generated correctly.

### Testing I've completed

I've installed the application using this script and checked that OpenSim is appearing in the applications menu and working.

### CHANGELOG.md (choose one)

- updated.


aymanhab:
Thanks for fixing this @AlbertoCasasOrtiz Considering that we don't distribute and don't want to give impression that we distribute the application on ubuntu I'd remove the change in changelog. Not sure if/how this worked in ubuntu 18 which is built by ci, but if we're possibly supporting ubuntu 20 I'd add that to ci as well. Thank you.

AlbertoCasasOrtiz:
Since the original repository from where this pull request was made has been deleted, I close this pull request. I have created another one (#1392) with the same changes, but without updates in the CHANGELOG.md file.

PR #1387, opened 2022-10-04
Build on osx11 since 10.15 is retiring on github actions

aymanhab:
Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated...


aymanhab:
@carmichaelong Can you check the GUI build on Mac osx 11 if it runs on your Mac? Please mention any other locals who may have a Mac. Other platforms are not affected by this PR and Mac 10.15 artifact was not uploaded to avoid overwriting/corruption.
For me it failed to launch on my ancient Mac 10.15 due to missing libgcc which may be expected.

carmichaelong:
Same issue in MacOS 12.6 (with an M1 chip). I've seen `libgcc_s.1.x` before in various places in CMake. A couple of places are located here:
- https://github.com/opensim-org/opensim-core/blob/524cb877dcea6d67ce7e68ad994792623c80663a/Vendors/tropter/cmake/CMakeLists.txt
- https://github.com/opensim-org/opensim-core/blob/a01cf545485e8bc1f6e50b29ddeab229677de2c5/Vendors/tropter/cmake/tropter_install_mac_dependency_libraries.cmake.in

The file `tropter_install_mac_dependency_libraries.cmake.in` in particular catches my eye.

aymanhab:
@AlbertoCasasOrtiz This PR has pointers to which cmake related files need updating before we can claim mac-osx11 is supported on ci

Issue #1388, opened 2022-10-13
Ubuntu 18.04 and MacOS 10.15 deprecated in CI.


AlbertoCasasOrtiz:
### Steps to reproduce

Open any CI Workflow and check the Annotations section, just below the continuous-integration.yml section.

### Expected result

No warnings

### Actual result

The following warnings are appearing:

- The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002
- The macOS-10.15 environment is deprecated, consider switching to macos-11(macos-latest), macos-12 instead. For more details see https://github.com/actions/virtual-environments/issues/5583

# Solution

Update CI environment versions.

# Notes:

- Ubuntu 18.04 will be fully unsupported by 4/1/2023.
- MacOS 10.15 will be fully unsuported by 12/1/22.
- There are some periods in which the CI jobs executed will fail. Those periods are listed here:
  - buntu 18.04: https://github.com/actions/virtual-environments/issues/6002
  - MacOS 10.15: https://github.com/actions/virtual-environments/issues/5583

AlbertoCasasOrtiz:
@aymanhab I think this issue is high priority, so I will take care of it as soon as possible. Should I just update the deprecated environments, or should I add the six available environments to CI? That would include:

- Ubuntu 20.04
- Ubuntu 22.04
- macOS 11
- macOS 12
- Windows Server 2019
- Windows Server 2022

Thank you.

PR #1389, opened 2022-10-13
Updated deprecated environments in CI.

AlbertoCasasOrtiz:
Fixes issue #1388

### Brief summary of changes

Updated environment versions for Mac and Ubuntu to latest.

### Testing I've completed

Running CI job.

### CHANGELOG.md (choose one)

- no need to update because this is a CI matter.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -145,7 +145,7 @@ jobs:
   mac:
     name: Mac
 
-    runs-on: macos-latest
+    runs-on: macos-11
 
     steps:
     - name: Checkout opensim-gui
@@ -294,7 +294,7 @@ jobs:
   ubuntu:
     name: Ubuntu
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
 
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -145,7 +145,7 @@ jobs:
   mac:
     name: Mac
 
-    runs-on: macos-10.15
+    runs-on: macos-latest
 
     steps:
     - name: Checkout opensim-gui
@@ -294,7 +294,7 @@ jobs:
   ubuntu:
     name: Ubuntu
 
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
 
     steps:
 


aymanhab:
While it would be good to have builds with latest versions, there're deeper issues that need to be addressed on opensim-core side first (e.g. https://github.com/opensim-org/opensim-core/pull/3295).
Also, since we build our distributions on github and users may not have the latest environment according to github we need actions to continue building with lowest supported os versions available on github as well (e.g. Windows 2019, Osx-11 and ubuntu 20).

AlbertoCasasOrtiz:
@aymanhab As we talked about this morning, I have set the Ubuntu version and the MacOs version to the lowest versions supported by GitHub Actions.

aymanhab:
Thanks @AlbertoCasasOrtiz We need to check that the built application launches correctly without setting LD_LIBRARY_PATH or DYLD_LIBRARY_PATH on osx, otherwise we need to go back and fix since the application installer doesn't set these and asking users to do it is highly not recommended.

AlbertoCasasOrtiz:
@aymanhab I just checked and the generated artifact is throwing the "Failed to load one or more dynamic libraries" error for: libgcc_s.1.1.dylib.

aymanhab:
Thanks @AlbertoCasasOrtiz that's why we need to go back to simtk-core and fix the cmake hacks described here https://github.com/opensim-org/opensim-gui/pull/1387

AlbertoCasasOrtiz:
@aymanhab Looks like I didn't set the correct syntax for powershell/windows in CI for these lines:

https://github.com/opensim-org/opensim-gui/blob/bebac6984b17d84bc2dddb5e5fe452d51ae9f8f2/.github/workflows/continuous-integration.yml#L49

https://github.com/opensim-org/opensim-gui/blob/69c163606aa532e7751426eeb21ee02b874b599e/.github/workflows/continuous-integration.yml#L92

https://github.com/opensim-org/opensim-gui/blob/bebac6984b17d84bc2dddb5e5fe452d51ae9f8f2/.github/workflows/continuous-integration.yml#L131

I'd change it, but I'd prefer not to modify CI while you are working on it to avoid merging conflicts. Could you change `$GITHUB_ENV` by `$env:GITHUB_ENV`, and `$GITHUB_OUTPUT` by `$env:GITHUB_OUTPUT` in those lines, please?

AlbertoCasasOrtiz:
@aymanhab I think we can close without merging since the current CI is working properly

aymanhab:
Agreed, Thanks @AlbertoCasasOrtiz

Issue #1390, opened 2022-10-13
INSTALL script in Ubuntu only working in Ubuntu 18.04


AlbertoCasasOrtiz:
### Steps to reproduce

Follow [Build Instructions](https://github.com/opensim-org/opensim-gui/wiki/Build-Instructions) for Ubuntu.

### Expected result

A system with Ubuntu >18.04 should be able to complete the installation process.

### Actual result

An message appears indicating that the only supported Linux distribution is Ubuntu 18.04.

### Environment and GUI version
Environment: Ubuntu >18.04
GUI version: Latest in master.

AlbertoCasasOrtiz:
Fixed in #1391

PR #1391, opened 2022-10-13
Checking that Ubuntu version is higher or equal than 18.04, instead of exactly 18.04.

AlbertoCasasOrtiz:
Fixes issue #1390 

### Brief summary of changes

Instead of check if the version of the Ubuntu distribution is exactly 18.04, now the installer checks if the version is higher or equal to 18.04.

### Testing I've completed

Manually inserted different versions of the Ubuntu in the script, to check if the if statement works.

### CHANGELOG.md (choose one)

- no need to update because we don't distribute the application on Ubuntu.

### DIFF
--- a/Gui/opensim/installer_files/Linux/INSTALL
+++ b/Gui/opensim/installer_files/Linux/INSTALL
@@ -2,7 +2,7 @@
 
 function err_unsupported_env {
     printf "\r  [\033[0;31mERROR:\033[0m] Your distro/release is not supported by this script\n"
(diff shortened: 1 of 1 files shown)


AlbertoCasasOrtiz:
Yes, currently we are checking if the OS is specifically Ubuntu. However, it should work for any Debian-based OS.

I think the minimal version of Debian that could work is Debian +10 (this is the version on which Ubuntu 18.04 is based), but this is something we have to test.

PR #1392, opened 2022-10-13
Fixed desktop icon in Ubuntu.

AlbertoCasasOrtiz:
Fixes issue #1373. This pull request substitutes #1374, since the original repository from which that pull request was made has been deleted.

### Brief summary of changes

Changed INSTALL script so the opensim.desktop file is generated correctly.

### Testing I've completed

I've installed the application using this script and checked that OpenSim is appearing in the applications menu and working.

### CHANGELOG.md (choose one)

- no need to update because we don't distribute the application on Ubuntu.


aymanhab:
@AlbertoCasasOrtiz Are we using opensim lower case across the board on linux? if so and the functionality works feel free to merge since we have barely any other users on linux/ubuntu. Thank you

AlbertoCasasOrtiz:
Yes, the opensim script used to invoke OpenSim in Ubuntu is lower case. I tested this on 18.04 20.04 and 22.04 a while ago.

Issue #1395, opened 2022-10-24
Marker selection in Navigator is not synced with Properties


tkuchida:
### Steps to reproduce
- open arm26 model
- in Navigator, expand "Markers"
- double-click "r_acromion" in Navigator
- double-click "r_humerus_epicondyle" in Navigator

### Expected result
The "r_humerus_epicondyle" marker is highlighted in the Navigator and selected in the Visualizer, and the Properties window displays its properties.

### Actual result
The "r_humerus_epicondyle" marker is highlighted in the Navigator and selected in the Visualizer, but the Properties window is showing the properties for "r_acromion". This should not be possible.

![](https://user-images.githubusercontent.com/4203505/197648732-38785b25-1600-42ad-8df9-565579fc1008.png)

### Environment and GUI version
Public release of 4.4

aymanhab:
Thanks for reporting @tkuchida Indeed I reproduced it on Windows. Single click however works as expected but I'm investigating further. Can you confirm what platform was that reported on? Thanks

aymanhab:
Notes from investigation so-far: 
Selecting visually in 3D view (1 click) results in all the correct updates in navigator, properties
Single select in navigator, updates properties window but doesn't select in 3D view
Double-click in navigator highlights in 3D view correctly but fails/incorrect in Properties window

tkuchida:
> Can you confirm what platform was that reported on?

Windows 10 Pro; installed via the "official release" Windows installer on simtk 🐴

PR #1401, opened 2022-11-08
Fix marker selection 1395

aymanhab:
Fixes issue #1395
### Brief summary of changes
Instead of navigator observing and making node selection, rely on netbeans plumbing to change/update selected nodes

### CHANGELOG.md (choose one)
- updated...


aymanhab:
I tested the behavior in issue #1395 and it is fixed in the artifact on Windows. Would be good for another user to test/verify @tkuchida or @AlbertoCasasOrtiz or @carmichaelong before we merge. The change is minimal but a bit complicated to explain.

AlbertoCasasOrtiz:
@aymanhab Tested on Windows 10 and Ubuntu 22.04. Looks good to me.

aymanhab:
Thanks @AlbertoCasasOrtiz Merging after updating changelog.

Issue #1378, opened 2022-07-20
In Static Optimization Tool, dialog box has wrong title for "Directory" in output section


carmichaelong:
### Steps to reproduce
1. Open a model
2. Open Static Optimization Tool
3. Click the folder icon under Output -> Directory

### Expected result
Title says something like "Choose output directory"

### Actual result
Title of dialog box says "Opensim model (.osim)" (or sometimes some other title). Also, the file type chooser shows .osim, but it should highlight that a directory should be chosen.

### Environment and GUI version
Mac, OpenSim 4.4 beta

PR #1402, opened 2022-11-16
Fixed bug in FileChooser for output directories.

AlbertoCasasOrtiz:
Fixes issue #1378 

### Brief summary of changes

When opening FileChooser to select the Output Directory on Analysis Tools, it was not updating the title and the filters. Thus, it was showing the title and filters set the last time you opened the FileChooser otherwhere (This is the reason the title and the filters were changing over time, it depends on the value assigned last time you opened a FileChooser). This has been solved by:

 - Added function to allow set title of the window without having to add elements to the FileFilter.
 - Reset FileFilter when the FileChooser is opened.
 - Make sure that the FileChooser is in save mode.

### Testing I've completed

Locally tested on Ubuntu 22.04.

### CHANGELOG.md (choose one)

- Updated.

### DIFF
--- a/Gui/opensim/utils/src/org/opensim/swingui/FileTextFieldAndChooser.java
+++ b/Gui/opensim/utils/src/org/opensim/swingui/FileTextFieldAndChooser.java
@@ -43,7 +43,7 @@ public class FileTextFieldAndChooser extends javax.swing.JPanel implements Actio
 
    private FileFilter filter = null;
    private boolean directoriesOnly = false;
-
+   private String title;
    private String lastFileName; // keep track of previous value to avoid firing change events when name stays the same
    private boolean fileIsValid = true;
    private boolean treatUnassignedAsEmptyString = true;
@@ -83,6 +83,10 @@ public class FileTextFieldAndChooser extends javax.swing.JPanel implements Actio
 
    public void setFileFilter(FileFilter filter) {
       this.filter = filter;
+   }
+	
+   public void setDialogTitle(String title) {
+	  this.title = title;   
    }
 
    public void setIncludeOpenButton(boolean includeOpenButton) {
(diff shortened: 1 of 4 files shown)


aymanhab:
Thanks @AlbertoCasasOrtiz Please resolve conflicts, the change looks good. Thank you.

Gui/opensim/utils/src/org/opensim/swingui/FileTextFieldAndChooser.java
```diff
@@ -84,6 +84,10 @@ public void setExtensionsAndDescription(String extensions, String description) {
    public void setFileFilter(FileFilter filter) {
       this.filter = filter;
    }
+	
+   public void setDialogTitle(String title) {
+	  this.title = title;   
+   }
```
aymanhab, line 91: Are there scenarios where title wouldn't be cleared from one invocation to the next?

Gui/opensim/utils/src/org/opensim/utils/FileUtils.java
```diff
@@ -262,6 +262,8 @@ public String browseForFolder(Frame parent, String description, boolean allowCre
         dlog.setCurrentDirectory(new File(defaultDir));
         dlog.setFileSelectionMode(JFileChooser.DIRECTORIES_ONLY);
         
+		dlog.resetChoosableFileFilters();
+
```
aymanhab, line 266: Please fix indentation

Gui/opensim/utils/src/org/opensim/swingui/FileTextFieldAndChooser.java
AlbertoCasasOrtiz: I have tested and it always sets the new title. I don't think there is any scenario where it couldn't be cleared.

aymanhab (approved):
Thanks @AlbertoCasasOrtiz

PR #1406, opened 2022-11-30
Revised INSTALL script, fix #1394.

### LINKED ISSUE #1394: Ubuntu group not being retrieved properly in INSTALL script
Found by an user in [this](https://github.com/opensim-org/opensim-gui/issues/1373#issuecomment-1273018917) comment.
More details in [this](https://github.com/opensim-org/opensim-gui/issues/1373#issuecomment-1281310728) comment.
### Steps to reproduce

Follow [Install instructions for Ubuntu](https://github.com/opensim-org/opensim-gui/wiki/Build-Instructions) until `bash INSTALL` is reached.

Your Ubuntu user must have a different name from the Ubuntu group it belongs.

### Expected result

INSTALL script finish successfully.

### Actual result

The following error message appears:
`chown: invalid group: ‘<user-name>:<group-name>’`

**Environment:** Ubuntu 20.04
**GUI version:** Code retrieved from master using [build script](https://github.com/opensim-org/opensim-gui#building-on-linux-beta) in README.

AlbertoCasasOrtiz:
Fixes issue #1394

### Brief summary of changes

 - Removed duplicated code.
 - Fixed problem with minimum version checking.
 - Fixed problem with user/group retrieval.

### Testing I've completed

- Tested on Ubuntu and Debian.

### CHANGELOG.md

- No need to update because we don't officially support Linux.

### DIFF
--- a/Gui/opensim/installer_files/Linux/INSTALL
+++ b/Gui/opensim/installer_files/Linux/INSTALL
@@ -90,55 +90,24 @@ else
     err_unsupported_env
 fi
 
-MIN_UBUNTU_VER="18.04"
+MIN_UBUNTU_VER="18"
 MIN_DEBIAN_VER="10"
 
-if [[ $OS == *"Ubuntu"* ]] && [[ $(echo $VER'>='$MIN_UBUNTU_VER | bc -l) ]]; then
+# Retrieve group of current user.
+GROUP=$(id -gn)
 
-    [[ $v == 1 ]] && echo "Installing Java..."
-    sudo apt install -qqqq -y openjdk-8-jre
+VER_SPLIT=(${VER//./ })
+VER_MAJOR=${VER_SPLIT[0]}
 
-    [[ $v == 1 ]] && echo "Testing for system LAPACK/BLAS..."
-    if ! (ldconfig -p | grep -q "libblas|liblapack"); then
-        [[ $v == 1 ]] && echo "  Installing liblapack3..."
-        sudo apt install -qqqq -y liblapack3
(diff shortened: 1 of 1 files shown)


aymanhab:
@AlbertoCasasOrtiz Trying to use the script to fix an issue on the sprint I ran into these 2 problems:
1. Requiring Ninja seems too strict especially that Ninja likely ends up using Make underneath
2. The swig built fails in CMake because of libpcre.so.1 is not found, maybe it needs to be installed first.

AlbertoCasasOrtiz:
@aymanhab I have tested the Linux script in a clean Ubuntu 22.04 and it works for me. `libpcre3` and `libpcre3-dev` are installed in line [Line 67](https://gist.github.com/AlbertoCasasOrtiz/6e4dce4cb8448e4989e33dae741df948#file-opensim-linux-build-script-sh-L67). What OS version are you using? Maybe I missed something when testing it.

Regarding ninja, I agree. Should we go back to cmake?

aymanhab:
I'm using 22.04.1 LTS according to lsb_release -a.
I'd use ninja if available otherwise use make. Goal is that internal dev. builds and ci builds are identical so issues/can be detected as early as possible. I'll look into build log to find out about pcre issues

AlbertoCasasOrtiz:
@aymanhab I think we can merge this PR since it is related to the file

https://github.com/opensim-org/opensim-gui/blob/c60b07fa6ed992e2221af2050daac74d67d6e62c/Gui/opensim/installer_files/Linux/INSTALL

and not to the build scripts. I tested this on Ubuntu and Debian and it works.

Issue #1412, opened 2022-12-14
Missing Tools in Tools Menu


AlbertoCasasOrtiz:
### Steps to reproduce

Install OpenSim-GUI, either using the build scripts or from an artifact installer, and check if all of the tools are available in the menu. More information in #1405.

### Expected result

All of the tools are available in the menu.

### Actual result

Some tools are missing. Only **Scale Model...**, **IMU Placer...**, **Inverse Kinematics...**, and **IMU Inverse Kinematics...** are available.

<img src="https://user-images.githubusercontent.com/15672785/207708533-0d120879-3726-4bfe-afb7-ce27c5b64e5b.png" width=400></img>

### Environment and GUI version

Problem replicated in Windows, Debian and Ubuntu, using artifact and builds generated from master branch.

AlbertoCasasOrtiz:
CI generated build from Dec, 7 was working. https://github.com/opensim-org/opensim-gui/actions/runs/3644659114
CI generated build from Dec, 8 was not working. https://github.com/opensim-org/opensim-gui/actions/runs/3653889015

Between these two days, the only change is this commit I made related to Issue #1357.

https://github.com/opensim-org/opensim-gui/commit/913b3a2946dc40d12a138dbb562cd6f9905b25fc

aymanhab:
Thanks for investigating @AlbertoCasasOrtiz and the pointer to the likely change that caused it, I'll fix

aymanhab:
I reverted the change from the commit above and reopened the issue pending fix so others are not on hold.

AlbertoCasasOrtiz:
@aymanhab I think I have solved it in #1414. Is there any way of automating GUI testing? I have seen this issue regarding that #328.

aymanhab:
We'd love to automate GUI testing as you found on #328 but that seems like a significant undertaking that we can investigate after the current sprint and decide where it fits in our development priorities.

aymanhab:
@AlbertoCasasOrtiz Let's refactor the two PRs to fix toolbar and restore missing commands into one PR, let ci build it and we can test the integrated artifacts on windows, mac and linux, if all works we can merge and close both issues.

PR #1413, opened 2022-12-14
Revert fix to 1357 that caused some tools to disappear

aymanhab:
Fixes issue #1412

### Brief summary of changes
Revert changes in PR fixing 1357 that had the side effect of disappearing some tools from ci builds
### Testing I've completed
Local build has all tools as expected
### CHANGELOG.md (choose one)

- updated...


AlbertoCasasOrtiz:
The revert looks good to me. I think I have solved the problem in #1414. If that PR does not work, we should proceed with the revert.

PR #1414, opened 2022-12-14
Fixed paths to missing tools. Fix #1412.

### LINKED ISSUE #1412: Missing Tools in Tools Menu

AlbertoCasasOrtiz:
Fixes issue #1412 

### Brief summary of changes

In 913b3a2946dc40d12a138dbb562cd6f9905b25fc, I changed the toolbar from *Edit* to *UndoRedo*, since *UndoRedo* automatically hides when the window is too small (when the toolbar was *Edit*, it was not visible/usable). However, I forgot to change the paths of the Tools. I have changed the paths here and it works now.

### Testing I've completed

Local build, checking if the tools are available.
We should make sure nothing else was broken in that commit.

### CHANGELOG.md (choose one)

- Updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ v4.4.1
 ======
 - Fix issue [#1378](https://github.com/opensim-org/opensim-gui/issues/1378): In Static Optimization Tool, dialog box has wrong title for "Directory" in output section.
 - Fix issue [#1395](https://github.com/opensim-org/opensim-gui/issues/1395) where visual selection and properties windows are not in sync.
-
+- Fix issue [#1412](https://github.com/opensim-org/opensim-gui/issues/1412): Missing Tools in Tools Menu.
 
 v4.4
 ====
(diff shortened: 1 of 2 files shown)


aymanhab:
Thanks @AlbertoCasasOrtiz, because we'll need to test the artifacts on the various platforms manually (see how it handles layout, resizing etc.) I lean to undo the previous PR fully 
https://github.com/opensim-org/opensim-gui/pull/1413
then introduce a new one that includes this fix that we test extensively before merging. 
There'd be no changelog entry for this potential fix to a fix that users never saw anyway.

AlbertoCasasOrtiz:
@aymanhab I agree, let's revert and merge this only once we are sure it is working properly.

aymanhab:
Awesome, thanks @AlbertoCasasOrtiz for the quick turnaround, will merge this PR once ci finishes to verify the tools show back as they should in the artifacts.

aymanhab:
The change looks good to me @AlbertoCasasOrtiz, thank you 👍 I would restructure though to include the original fix that was reverted so that we have one PR and a set of artifacts to test with on all platforms. No need to rush this until the current sprint tasks are done.

aymanhab:
This can now be closed without merging since it's folded into #1418

Issue #1420, opened 2023-01-19
INSTALL script not installing java in Debian


AlbertoCasasOrtiz:
### Steps to reproduce

Try to install opensim on a Debian environment without java.

### Expected result

Opensim is installed.

### Actual result

Opensim complaining about java not being located, and asking to set --java-path.

### Solution

Install java here if the OS is Debian:

https://github.com/opensim-org/opensim-gui/blob/4981b76f42edf33eaaeab9f49e6ac24dcc888ba5/Gui/opensim/installer_files/Linux/INSTALL#L106

Maybe by solving this issue we can create a single installer that can be used across different Debian-based OS.

### Environment and GUI version
Debian 10 and 11.
GUI in master.

PR #1421, opened 2023-01-23
 INSTALL script not installing java in Debian 

AlbertoCasasOrtiz:
Installing java for Debian in Linux installer.

Fixes issue #1420

### Brief summary of changes

The installer was not installing Java in Debian. Not it installs it.

### Testing I've completed

Tested locally in Debian 11.

### CHANGELOG.md (choose one)

- no need to update because this is an installer issue.


AlbertoCasasOrtiz:
I have tested this again and it is working in both Ubuntu and Debian.

Issue #1426, opened 2023-02-06
Selecting pathpoints in the 3D window for path editing is flaky


aymanhab:
### Steps to reproduce
Reported by Scott Delp and 281 student

### Expected result
Picking a pathpoint to edit, it gets highlighted and triad to move it is shown
### Actual result
Sometimes a different object in the vicinity is selected instead. Selecting from the path editing dialog always works (as a workaround)
(include screen grabs where appropriate)

### Environment and GUI version
4.4

aymanhab:
Looking into the visualizer code, there was change in the order in which objects are returned upon attempted selection from threejs version used in 4.3 to the latest, we need to handle it better by prioritizing objects that are advertised as "movable" in OpenSim (markers and pathpoints)

PR #1427, opened 2023-02-06
Include fix to pathpoint selection from threejs repo

aymanhab:
Fixes issue #1426 

### Brief summary of changes
On graphical selection/pick, prioritize pathpoints and markers as the only types graphically editable in application.

### Testing I've completed
Tutorial 2
### CHANGELOG.md (choose one)

-  Will update after testing and deciding whether to backport.


aymanhab:
If we can test this PR artifacts @carmichaelong that would be great. Not sure if worth backporting and building the GUI to include this. We can discuss in dev mtg.

PR #1428, opened 2023-02-07
update threejs with selection fix

aymanhab:
Fixes issue #1426

### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ v4.4.1
 - Fix issue [#1395](https://github.com/opensim-org/opensim-gui/issues/1395): Where visual selection and properties windows are not in sync.
 - Fix issue [#1357](https://github.com/opensim-org/opensim-gui/issues/1357): Toolbar displaced and non visible after resizing window
 - Fix issue [#1416](https://github.com/opensim-org/opensim-gui/issues/1416): Remove visual offset added to experimental data in "Preview Experimnetal Data".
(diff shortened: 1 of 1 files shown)


aymanhab:
@carmichaelong for your review/testing when you have a chance. Thanks

carmichaelong (approved):
@aymanhab Testing looked good on my end. One minor suggestion (i.e., not blocking) and one minor thing I ran into (but is not new).

**Suggestion**
In the code from the `threejs` repo:
```
var foundPickable = false;
for (let i = 0; i < intersects.length && !foundPickable; i ++) {
	var nextObject = intersects[ i ].object;
	foundPickable =  nextObject.opensimType == "PathPoint" || 
		nextObject.opensimType == "ConditionalPathPoint" || 
		nextObject.opensimType == "Marker";
	if (foundPickable)
		object = nextObject;
}
```
The boolean check on `foundPickable` in the for loop can be removed by adding a `break;` in the `if(foundPickable)` branch. Then, `foundPickable` also does not need to be declared before the for loop.

**Note**
While testing, I was clicking away from path points intentionally, however in many cases instead of getting the bone, I would pick a path point of a muscle that has been hidden. Checking the 4.4 release GUI, this is not a new issue.

aymanhab:
Thanks @carmichaelong while true that the variable could be removed, the current iteration made the testing of the control flow easier so I'll leave it as is. The fact that we can select hidden objects is rather surprising, I'd open an issue to see if we can handle it more gracefully.

PR #1432, opened 2023-03-07
Added build scripts

AlbertoCasasOrtiz:
Fixes issue https://github.com/opensim-org/opensim-core/issues/3397

### Brief summary of changes

Added build scripts.

### Testing I've completed

**build-scripts:** Tested in previous iterations of the scripts, now I am only copying them here.

### CHANGELOG.md (choose one)

- no need to update because internal scripts.


aymanhab (approved):
LGTM, thanks @AlbertoCasasOrtiz 
Not sure if the options for Moco or wrapping were exercised but GUI requires building with Java wrapping on and with ezc3d on otherwise will not compile

PR #1448, opened 2023-06-18
Update swig bindings to use swig 4.1.1

aymanhab:
Fixes issue #0

### Brief summary of changes
Regenerate java files through swig 4.1.1 which was forced by change in ci/opensim-core

### Testing I've completed
None other than building locally and spot checking.
We need to decide if we change documentation to require 4.1.1 for building or support both with the potential of diverging if features in 4.1.1 are used that were not available in 4.0.2.

### CHANGELOG.md (choose one)

- updated...


AlbertoCasasOrtiz:
@aymanhab I have tested the artifacts in windows and mac and it is working for me. The MacOs VM is not working, so I do not have a way of testing it in Mac.

AlbertoCasasOrtiz (approved):
LGTM, but I cannot review in Mac.

aymanhab:
Thanks @AlbertoCasasOrtiz merging so we can put mileage on artifacts on all platforms.

PR #1461, opened 2023-09-21
Update GUI codebase to handle GeometryPath refactoring

aymanhab:
Fixes issue #<0>

### Brief summary of changes
Update bindings to be consistent with opensim-core post GeometryPath refactoring

### Testing I've completed
Built and launched the GUI on windows and visualized a model with muscles and another without muscles

### CHANGELOG.md (choose one)


aymanhab:
Need to test the artifacts on windows and Mac before merging.

aymanhab:
@carmichaelong if you can test it would be great. Note that the ubuntu build due to out-of-space on ci build machine is unrelated and has been there since forever. Thank you

AlbertoCasasOrtiz:
@aymanhab I have tested on windows loading different models and everything seems to work properly.

carmichaelong:
Testing on both Windows and Mac, models seem to load OK, but either "Show"ing or "Hide"ing muscles does not seem to do anything (no error messages either)

aymanhab:
Thanks @carmichaelong as it turns out the GUI code was counting on Property name being GeometryPath in a couple places, now changed to use "path" instead and utilize the hasVisualPath() hack. Will need another round when we have actual models with non GometryPath paths but for now this fix handles the show/hide/edit as far as I can tell.  You can review now or when ci is finished.

Gui/opensim/view/src/org/opensim/view/nodes/OneForceNode.java
```diff
@@ -55,9 +57,10 @@ public class OneForceNode extends DisablablModelComponentNode implements Colorab
      */
     public OneForceNode(OpenSimObject force) {
         super(force);
-        hasPath = force.hasProperty("GeometryPath");
+        Force typedForce = Force.safeDownCast(force);
+        hasPath = typedForce.hasProperty("path") && typedForce.hasVisualPath();
```
carmichaelong, line 61: Is it sufficient to just use `typedForce.hasVisualPath()` here?

carmichaelong (commented):
@aymanhab thanks, just one quick comment in the code, and then I can test after that's resolved

Gui/opensim/view/src/org/opensim/view/nodes/OneForceNode.java
aymanhab: ### Based on the comments in hasVisualPath I believe you're right. Thanks for the review @carmichaelong

carmichaelong (approved):

carmichaelong:
LGTM, tested with the Mac artifact

aymanhab:
Thanks @carmichaelong , @AlbertoCasasOrtiz  considering testing same behavior on windows by myself, will merge.

PR #1464, opened 2023-10-30
Update bindings

aymanhab:
Fixes issue #0

### CHANGELOG.md (choose one)

- updated...


aymanhab:
@nickbianco For your review. Typically we only see a few file changes due to added/removed methods in the API, this time we got a ton of new files with prefix SWIGTYPE_p_* which are just clutter and typically pollute the bindings with unusable classes. Historically this happens when we change the ordering of include files. Wanted to bounce this off you since the last change before this with refactoring Geometry Path and it didn't have this side-effect. Thank you.

PR #1471, opened 2023-11-15
Update to latest opensim-models repo

aymanhab:
Fixes issue #0

### Brief summary of changes
Use latest opensim-models repo.

This is a place holder, If we include new Rajagopal model with muscles then we should get it in first .

### Testing I've completed
None
### CHANGELOG.md (choose one)

- Typically gui issues only are included here


nickbianco:
@aymanhab I've added an updated version of the Jupyter notebooks to `opensim-core`: https://github.com/opensim-org/opensim-core/tree/main/Bindings/Python/tutorials. These tutorials should be included in the OpenSim 4.6 distribution. I don't have any models to add at the moment.

PR #1486, opened 2024-02-17
Workflow to run on PR from opensim-core, builds on windows only

aymanhab:
Fixes issue #0
### Brief summary of changes
Separate workflow file to be invoked when a PR going into opensim-core is submitted.

### CHANGELOG.md (choose one)


aymanhab:
Since this has no code change and may not run if not on main branch, will merge then update/test after it is in place

Issue #1494, opened 2024-03-10
Error when changing muscle color


mrrezaie:
### Steps to reproduce
1. open a model with muscle(s) (e.g. Rajagopal)
2. select muscle(s) from navigator (Forces>Muscles>all)
3. right click on the selected muscle(s) and select Display>Color...
4. pick a color and hit OK
### Expected result
The color of the selected muscle(s) should change.
### Actual result
Raises exception(s):
```
java.lang.RuntimeException: Property 'GeometryPath' not present in Object addbrev_r
	at org.opensim.modeling.opensimCommonJNI.OpenSimObject_getPropertyByName(Native Method)
	at org.opensim.modeling.OpenSimObject.getPropertyByName(OpenSimObject.java:328)
	at org.opensim.threejs.ModelVisualizationJson.propagateGeometryPathCommandsToPathPoints(ModelVisualizationJson.java:1410)
	at org.opensim.view.pub.ViewDB.applyPendingAppearanceChanges(ViewDB.java:219)
...
```
```
java.lang.NullPointerException
	at org.opensim.view.pub.ViewDB.applyPendingAppearanceChanges(ViewDB.java:214)
	at org.opensim.view.pub.ViewDB.setApplyAppearanceChange(ViewDB.java:199)
	at org.opensim.view.ObjectDisplayColorAction.performAction(ObjectDisplayColorAction.java:53)
	at org.openide.util.actions.CallableSystemAction$1.run(CallableSystemAction.java:105)
	at org.openide.util.actions.ActionInvoker$1.run(ActionInvoker.java:70)
	at org.openide.util.actions.ActionInvoker.doPerformAction(ActionInvoker.java:91)
	at org.openide.util.actions.ActionInvoker.invokeAction(ActionInvoker.java:74)
	at org.openide.util.actions.CallableSystemAction.actionPerformed(CallableSystemAction.java:102)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.AbstractButton.doClick(AbstractButton.java:376)
	at javax.swing.plaf.basic.BasicMenuItemUI.doClick(BasicMenuItemUI.java:882)
[...]

PR #1495, opened 2024-03-10
Handle property name change from GeometryPath to path in visualization

aymanhab:
Fixes issue #1494 

### Brief summary of changes
Missed  spot when replacing Property GeometryPath changing name to "path" in 4.5 stream.

### CHANGELOG.md (choose one)

- updated...


carmichaelong (approved):
LGTM. Tested this artifact on Mac. Reproduced error in previous version by trying to change the color of one muscle in the `arm26` model, and this worked as expected in this artifact.

Issue #1496, opened 2024-03-13
Error when plotting MOT/STO files


mrrezaie:
### Steps to reproduce
1. open Plotter
2. load an MOT or STO file (IK or SO output in my case)
3. choose Y and X quantities
4. hit Add
### Expected result
I should see the curve.
### Actual result
Raises this error:
```
java.lang.RuntimeException: Array index out of bounds
	Thrown at Array.h:391 in get().
	at org.opensim.modeling.opensimCommonJNI.ArrayDouble_getitem(Native Method)
	at org.opensim.modeling.ArrayDouble.getitem(ArrayDouble.java:238)
	...
```

### Environment and GUI version
Windows 10, OpenSim 4.5-2024-03-11-765d625

aymanhab:
@mrrezaie Thanks for reporting, but please provide the specific file that caused the crash as it was not possible to reproduce with a generic .mot file locally.  Thank you

mrrezaie:
Hi, sorry, I was testing with this artifact https://github.com/opensim-org/opensim-gui/actions/runs/8228001907

mrrezaie:
@aymanhab is it because that build (artifact) has not been merged to the main branch yet?

mrrezaie:
@aymanhab, I tested the corresponding artifact and the Plotter worked well. Thanks for the fix. There is also another similar issue, so I'm posting it here:

In that [artifact](https://github.com/opensim-org/opensim-gui/actions/runs/8331332107), when saving and re-loading a setup file in Analyze tool, this happens to the actuators files box:
![image](https://github.com/opensim-org/opensim-gui/assets/38867713/90ed7e0f-4646-4f40-9175-e4eaa1f13863)
as well as this pop-up error:
`Object: Cannot construct from empty filename. No filename specified.
	Thrown at Object.cpp:98 in Object().`
This works correctly in the official 4.5. Thank you.

PR #1498, opened 2024-03-18
Replace Array.rfindIndex with findIndex to work around bug with Array

aymanhab:
Fixes issue #1496 
### Brief summary of changes

The new Array implementation returns incorrect/different index for rfindIndex than the legacy code. While the call site has been updated the funtion rfindIndex should be fixed for backward compatibility for possible use in client code elsewhere.

### Testing I've completed
Plots in the GUI

### CHANGELOG.md (choose one)

- Will update after core issue is resolved


nickbianco (approved):

aymanhab:
This was a workaround for a bug introduced on core, this has been fixed there so no need for the workaround. Will test artifact and close if issue is fixed without this PR

PR #1508, opened 2024-05-23
Update continuous-integration.yml

aymanhab:
Fixes issue #0

### Brief summary of changes
Build gui that include fix to opensim-core PR #3789

### CHANGELOG.md (choose one)

- updated...


aymanhab:
this was created for the side-effect  of building GUI against a specific opensim-core branch, there's another one for that now

PR #1511, opened 2024-05-30
Add function call from shell to recover camera parameters for scripting

aymanhab:
Fixes issue #0

### CHANGELOG.md (choose one)


aymanhab:
unnecessary now that we can save/recover cameras from json

PR #1524, opened 2024-10-04
Add locale requirements to linux build script

ksomml:
Fixes issue https://github.com/opensim-org/opensim-core/issues/3924

### Brief summary of changes

Added locale changes into the build script so numeric formatting is correct when using TimeSeriesTable.h

### Testing I've completed

Tested locally with a separate shell script. A restart after locale changes is recommended if locale changes have been applied due to different settings instead of directly using OpenSim after installation.

### Looking for feedback on...

/

### CHANGELOG.md (choose one)

- no need to update because the change only affects the build script

### DIFF
--- a/scripts/build/opensim-gui-linux-build-script.sh
+++ b/scripts/build/opensim-gui-linux-build-script.sh
@@ -82,6 +82,15 @@ else
 fi
 echo
 
+# Set UTF-8 locale and en_US locale for numeric formatting.
+echo "LOG: CHECKING AND SETTING LOCALE..."
+sudo apt update && sudo apt install --yes locales
+sudo locale-gen en_US.UTF-8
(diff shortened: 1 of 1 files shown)


scripts/build/opensim-gui-linux-build-script.sh
```diff
@@ -217,4 +226,4 @@ echo
 echo "LOG: INSTALLING OPENSIM-GUI..."
 cd ~/opensim-workspace/opensim-gui-source/Gui/opensim/dist/installer/opensim
 bash INSTALL
-echo
\ No newline at end of file
+echo
```
ksomml, line 229: This line change can be removed.

ksomml:
Closed due to feedback on
https://github.com/opensim-org/opensim-core/pull/3934

PR #1525, opened 2024-10-08
Update linux build script

ksomml:
Fixes issue with building OpenSim GUI on Linux.

### Brief summary of changes

Updated the build script to the latest changes inside the [opensim-core linux build script](https://github.com/opensim-org/opensim-core/blob/main/scripts/build/opensim-core-linux-build-script.sh)

### Testing I've completed

From my testing the OpenSim GUI 4.4 Release generally does not successfully build with any version of this build script (before and after modifications). Only the main branch seems to properly build the GUI. The only way I managed to install Core+GUI is to set the branch for the core to `4.5.1` and the branch for the gui to `main`.

### CHANGELOG.md (choose one)

- no need to update because I believe generally there should be more work done on improving these build scripts.

Suggestions on improving:
- **open a new repository just for build scripts and link them in opensim-core and opensim-gui** (very important in my opinion)
- give the opportunity to decide if core or GUI or both should be installed
- when installing opensim-core it should follow the standard linux install directory instead of installing it into the home directory
- instead of cloning the main branch by default (as it is at the moment) it should always clone the latest stable release instead
- instead of always downloading netbeans and cmake for example, it should check if it already exists and then decide to skip installation or not as for example netbeans will always be downloaded again which takes a long time due to the websites download throttle

### DIFF
--- a/scripts/build/opensim-gui-linux-build-script.sh
+++ b/scripts/build/opensim-gui-linux-build-script.sh
@@ -84,8 +84,26 @@ echo
 
 # Install dependencies from package manager.
 echo "LOG: INSTALLING DEPENDENCIES..."
-sudo apt-get update && sudo apt-get install --yes build-essential cmake autotools-dev autoconf pkg-config automake libopenblas-dev liblapack-dev freeglut3-dev libxi-dev libxmu-dev doxygen python3 python3-dev python3-numpy python3-setuptools libpcre3 libpcre3-dev libpcre2-dev byacc git gfortran libtool libssl-dev libffi-dev ninja-build patchelf || ( echo "Installation of dependencies using apt-get failed." && exit )
+sudo apt-get update && sudo apt-get install --yes build-essential cmake autotools-dev autoconf pkg-config automake libopenblas-dev liblapack-dev freeglut3-dev libxi-dev libxmu-dev doxygen python3 python3-dev python3-numpy python3-setuptools git libssl-dev libpcre3 libpcre3-dev libpcre2-dev libtool gfortran ninja-build patchelf libffi-dev byacc || ( echo "Installation of dependencies using apt-get failed." && exit )
 echo
+
+# If byacc available, install it (issue #3578)
+if apt-cache show byacc &> /dev/null; then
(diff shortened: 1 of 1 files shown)


scripts/build/opensim-gui-linux-build-script.sh
```diff
@@ -190,10 +208,24 @@ echo
...
+cd ~/opensim-workspace/opensim-core-build
+# TODO: Temporary for python to find Simbody libraries.
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:~/opensim-workspace/opensim-core-dependencies-install/simbody/lib
+ctest --parallel $NUM_JOBS --output-on-failure
+
+# Install opensim-core.
+echo "LOG: INSTALL OPENSIM-CORE..."
+cd ~/opensim-workspace/opensim-core-build
 cmake --install .
+echo 'export PATH=~/opensim-core/bin:$PATH' >> ~/.bashrc
```
halleysfifthinc, line 227: I'd recommend invoking the opensim install script instead.

halleysfifthinc (approved):

scripts/build/opensim-gui-linux-build-script.sh
halleysfifthinc: Hmmm actually I'm not sure that whole block (lines 223-229) is necessary. The opensim-core build outputs are copied/installed with the OpenSim GUI by default, and the bash script at https://github.com/opensim-org/opensim-gui/blob/220fe774fa69aa4090f2c5eb20d661e71d499ab8/scripts/build/opensim-gui-linux-build-script.sh#L251 calls the `./opensim-install-command-line.sh` script that installs a link to `opensim-cmd` in `/usr/local/bin`. I think as it stands, this build/install script is double installing opensim-core (one copy at the default install prefix, and another copy within the opensim GUI directory).

ksomml: That could be the case. I haven't had a look into that INSTALL script yet and won't be able to until Friday. Generally I hope this PR could be used to get some attention on the overdue refactor of the linux build scripts.

PR #1549, opened 2025-04-24
build tag 4.5.2 GUI

aymanhab:
Fixes issue #0

### Brief summary of changes
This has no code change, only checks out and builds core for tag 4.5.2 and build the stack cross platform in ci

### CHANGELOG.md (choose one)

- updated...


aymanhab:
This need not be merged at all, it's just an action to create GUI consistent with 4.5.2 tag of the opensim-core repo in case it's needed for RehabWeek @nickbianco @carmichaelong @mattpetrucci

Issue #1555, opened 2025-07-17
GUI crashes after IMU IK is finished


ClayZhang999:
Dear OpenSim team,

The virtual office hour is really helpful as before. And one thing I forgot to mention was the GUI crashing issue. 

I have seen some other users posting the same crashing issues in the forum too, as detailed via the link below. 

[https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=20639&p=0&start=0&view=&sid=78a77df0d6dc6d47b2b1f2ca09f37046](url)

I thought the Github issues would be more closely monitored, so I'm opening an issue here.

Many thanks,
Zhenyuan

aymanhab:
This was the result of a crash deleting the tool due to double deletion of OrientationWeightSet. Investigating

aymanhab:
This can be closed as fixed

PR #1559, opened 2025-09-11
Fix imuiktool memmgmt

aymanhab:
Fixes issue #1555
### Brief summary of changes

Changed an AdoptAndAppend set call to CloneAndAppend so that object created on the GUI side are not also owned by the native side.

### Testing I've completed
Issue 1555 doesn't reproduce with the fix. Also consistent with other reported crashes that happen due to garbage collection after running IMU-IKTool

### CHANGELOG.md (choose one)

- Will update before merging


carmichaelong (dismissed):
LGTM. Looks like Mac failure is unrelated (and not sure if it needs to be addressed in another PR or if things are changing with the new viewer soon).

aymanhab:
Thanks @carmichaelong the mac failure is unrelated and is caused by cmake version which has been addressed on the viewer branch so will be fixed when that is merged. Thank you 👍

PR #1577, opened 2025-12-17
Fix typos `contraint`

mrrezaie:
Fixes issue https://github.com/opensim-org/opensim-core/issues/4212

### Brief summary of changes
Fix the `contraint` typos

### Testing I've completed
None, will see the CI. It might fail due to cache.

### CHANGELOG.md (choose one)
- no need to update because it's trivial


aymanhab:
These are good changes but these files are auto generated, fix need to be in c++ side if not already. Thank you

PR #1601, opened 2026-03-10
Update viewer commit

aymanhab:
Fixes issue #0

### Brief summary of changes
Gather branding changes and latest viewer fixes in one artifact to test cross platform

### CHANGELOG.md (choose one)

- updated...


AlbertoCasasOrtiz:
Splash screen looking great.

After doing more testing, video recording shows an encoding error in my other computer, which means we may be capturing the error that caused the 0 bytes videos in the past weeks. I will work on it.

PR #1614, opened 2026-03-27
Fix viewer looping behavior with multiple motions

aymanhab:
Fixes issue #1616 
### Brief summary of changes
Loop button now force loop of all current animations  in the viewer, plus misc. fixes to issues that came up during beta testing of 4.6

### CHANGELOG.md (choose one)


aymanhab:
@AlbertoCasasOrtiz Looping behavior should be fixed now, looking into the non zero start time.

AlbertoCasasOrtiz:
Looks like it is working as expected now.

It also seems to have fixed the recording issue.

aymanhab:
Thanks @AlbertoCasasOrtiz Time to assess looping and recording while syncing and with non zero start time. Should be fixed now.

nickbianco:
@aymanhab I just tested these changes with artifact `OpenSim-4.6-2026-04-02-194f257-mac`, and I don't think the issue regarding recording and looping that I've encountered hasn't been fixed yet.

Here's a recording of an motion generated from IK. (Note that when I click record, the motion is not animating in the visualizer).

https://github.com/user-attachments/assets/512e861c-a11f-4f3e-b970-cdd07bff8ea4

Now, if I have an animation active and looping the visualizer, and _then_ I click record, I get the following video:

https://github.com/user-attachments/assets/a8b686e3-b7b2-48bb-b6e8-abd9ae2110af

Same length video, but the animation is played twice and at double speed. To me, it seems that the video recorder is manually advancing frames to record the motion, and when looping the animation, it's advancing frames that are already being advanced by the animation bar (if that makes sense).

Proposed solution: when recording videos, can we just take full control over the animation playback? So if someone is currently playing an animation (whether looping or not), the video recorder kills any looping, sets the animation back to the first time point, and animates forward at the normal playback speed.

aymanhab:
@nickbianco Thanks for testing 👍  Recording aside, is the behavior related to looping/syncing and non-zero start time fixed? Just to help decide if the issues are confined to the recording. Thank you

nickbianco:
It's not related to syncing and doesn't seem to be related to a non-zero start time. I don't think it's even necessarily related to looping.

If you have an animation running when you record a video, the recorder does not record a video at the correct speed.

nickbianco:
Here's what happen when I play an animation and then immediately hit "Record Video":

https://github.com/user-attachments/assets/5d5885b9-1d8f-47af-a37b-89536930a258

You can see when the animation stops playing: the video jumps back to it's normal speed.

aymanhab:
@nickbianco all the scenarios described in the bug report https://github.com/opensim-org/opensim-gui/issues/1616 did not mention recording. Can you verify if the issues reported there have been fixed? I'll look into the recording side.

aymanhab:
I agree a reasonable solution is to reset the animation to start and turn all controls off at the start of recording, at the end all the options will be enabled again.

nickbianco:
> @nickbianco all the scenarios described in the bug report https://github.com/opensim-org/opensim-gui/issues/1616 did not mention recording.

That issue is not related to what I'm discussing here. Apologies, I was just following up from our discussion during stand up and assumed this PR covered the recording issues too. If you'd prefer me to open a separate PR for the recording issues, let me know.

As for the issues described in #1616, the loop button issue I had before seems to be fixed, but there are still a couple of odd issues I noticed. I made a quick video summarizing the behavior:

https://github.com/user-attachments/assets/b164516b-2d3a-4438-9083-7261d99589d8

In short,
- If you try to reverse an animation while looping, it will try to loop around to the end of the motion, but seems to get stuck. You can see the model jittering as if it's trying to do something.
- A motion playing forward must be paused before played in reverse (and vice versa).

aymanhab:
Totally understand and im working on fixing the recording based on the above as I didn't touch it last time around. Thank you @nickbianco 🥇

aymanhab:
All should be fixed in the latest artifact. Doing more testing.

nickbianco:
Sounds good @aymanhab, let me know when you have an artifact ready to test.

aymanhab:
The artifacts are now ready to verify the fixes, I also shortened the prefix of default filenames we produce (opensim-viewer-video -> opensim-video) to make it fit in dialogs.

nickbianco:
Great! Giving it a try now.

nickbianco:
@aymanhab looking better, but I still found a few issues. The reverse playback options interact with playback and looping in unexpected ways.

- If you are playing a motion and have looping selected, clicking reverse play cancels the loop.
- However, you can start a reverse playback and _then_ click looping. When the animation reaches the initial time point it "tries" loop around to the final point, but the animation gets stuck.
- But, if you reverse frame-by-frame, looping does seem to work.

Here's another quick video summarizing the behavior:

https://github.com/user-attachments/assets/b41b1cd4-d240-435c-814d-d11d6b4a0ba2

aymanhab:
Thanks @nickbianco for testing 👍 

Generally speaking mixing reversing and looping do not make much sense and that's why it's not supported in the underlying threejs animations. There's a ping-pong mode where it goes to end then back but not loop backwards. As such, the behavior of choosing to reverse canceling looping is the desired behavior, maybe needs documentation but I would not spend time trying to make this rather unnatural mode work. I'd be interested if you have a use case.
As such I'd leave 1 as is, and disable the loop button while in reverse (thanks for the catch). 
Frame by frame is not actually an animation, it's just setting time to a specific value.

Any comments on recording behavior which was the meat of the comments earlier would be much appreciated.

aymanhab:
@AlbertoCasasOrtiz For your review, feedback when you have a chance. Thank you

aymanhab:
It may make sense to alert users to the new behavior by changing the icon for looping to include a forward arrow so it's not misleading..

AlbertoCasasOrtiz:
I just tested Tutorial 3 IK --> https://opensimconfluence.atlassian.net/wiki/spaces/OpenSim/pages/53089741/Tutorial+3+-+Scaling+Inverse+Kinematics+and+Inverse+Dynamics

And Observed a similar behaviour to what Nick reported. I Played the animation resulting from IK at 0.5, and, while it was running, clicked record.

It records only part of the animation, and then the last few frames are motionles in my case.

nickbianco:
@aymanhab sounds good. I'm fine with looping not working in reverse. Disabling the loop button while in reverse should fix the main issue. As Jen mentioned, it could be helpful to have some tutorial videos to help users with the new visualizer.

aymanhab:
@AlbertoCasasOrtiz I checked the same tutorial and you're right, likely something I missed in accounting for the non zero start time while recording. Are you able to troubleshoot? If not I will look into it, just let me know. Thank you

AlbertoCasasOrtiz:
I'm looking into it right now. I'll let you know what I find

aymanhab:
@AlbertoCasasOrtiz I believe I have fixed the issue by removing a line in the video-recorder that clamps time. Not sure what was it intended for but I committed the fix so you can assess once the artifacts finish building. Thank you

AlbertoCasasOrtiz:
At least the issue with Tutorial 3 seems fixed. I would wait for @nickbianco to confirm it works on Mac too.

nickbianco:
Testing now...

nickbianco:
@aymanhab, @AlbertoCasasOrtiz, seems to be mostly working. Here's a video that I recorded while an animation was playing, with speed set at 0.5:

https://github.com/user-attachments/assets/01b9a51a-2676-41d2-ae8e-a8c3353aab7a

The full animation is captured at the correct speed, which is good. However, the initial frame (as seen in the video's thumbnail) is not the first frame of the motion. Rather it is the frame that the model was in when I clicked record. We should eliminate this frame from video recording.

aymanhab:
@nickbianco Thank you very much for testing :+1: 
@AlbertoCasasOrtiz and/or I will look into the first frame issue.

aymanhab:
@nickbianco Can you explain how you produced this video? @AlbertoCasasOrtiz couldn't reproduce on Windows and my quick attempt to produce on mac also failed. Thank you.  
I found vlc to be very good at inspecting frames so indeed there's a wrong frame to start.

nickbianco:
I recorded the video with the viewer's video recorder as described above, i.e., I start recording a video in the middle of an animation. 

He's a video of me scrubbing through the video in both QuickTime and VLC. It seems like the initial frame is in the video itself (not just a thumbnail):

https://github.com/user-attachments/assets/fc941fc2-430d-41dc-b60e-28a1d7f717a8

aymanhab:
@AlbertoCasasOrtiz I fixed the issue with the slider being detached, only issue left is the extra frame. Did you have a chance to investigate? Thank you. If you can test the latest artifact to verify that would be awesome.

aymanhab:
@nickbianco Can you verify the slider grabbing, and first frame issue have been fixed in the latest artifact? I tested and it seems fixed but you're very good at reproducing these issues. Thank you

aymanhab:
@AlbertoCasasOrtiz Is there anything you plan to fix/modify based on the testing session? Thank you

AlbertoCasasOrtiz:
@aymanhab I don't think we need to modify anything specific. Aspect ratio will need changes but as we said will not be included in this version of opening GUI.

nickbianco:
@aymanhab just tested the latest artifact. The slider is now greyed out while animated, that looks great. And the first frame issue for video recoding looks fixed too. Nice job!

Issue #1648, opened 2026-05-27
Python examples not included with the OpenSim 4.6 beta installation


nickbianco:
Installed the OpenSim 4.6 beta (tag `OpenSim 4.6-2026-05-14-195f722`) and launched the GUI to install examples. Navigating to the Python examples directory (Code -> Python), the subfolders `Moco` and `PolynomialPathFitter` exist, but there are no Python files. (The Matlab directory contains the typical set of .m files).

nickbianco:
FYI @aymanhab

aymanhab:
Just installed on windows and all are there, will check osx, any specific example you're looking for that I maybe missing?

aymanhab:
This appears to be specific to osx, investigating

aymanhab:
Issue was ci script disabling build python wrapping which we don't distribute but that was getting in the way of copying the examples on osx. Should be fixed on ci here https://github.com/opensim-org/opensim-gui/actions/runs/26566248593#artifacts

PR #1652, opened 2026-06-02
Enable API examples build in CI configuration

aymanhab:
Fixes issue #1648 
### Brief summary of changes
Add BUILD_API_EXAMPLES=on to ci build scripts

### Testing I've completed
installed artifacts and they have the files at issue

### CHANGELOG.md (choose one)

- no need to update because internal.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -114,7 +114,7 @@ jobs:
         mkdir $env:GITHUB_WORKSPACE\\build_core
         chdir $env:GITHUB_WORKSPACE\\build_core
         $env:CXXFLAGS = "/W0 /MD"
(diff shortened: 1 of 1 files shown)


aymanhab:
I'd also like to point out the reason this flag was used  in corresponding cmakelists.txt.  It is incorrect to build actual API examples during the normal build process since they use/target an installation layout which is valid only after the build and install are finished. Conda builds which use Ninja underneath failed with this flag on because it attempted to build these examples prematurely. We're getting away with it due a fluke in the build system IMHO

nickbianco (approved):
LGTM.

nickbianco:
@aymanhab agreed. I've opened https://github.com/opensim-org/opensim-core/issues/4334 so that we can address the actual CMake issue in the next dev cycle.

aymanhab:
Awesome, thanks much @nickbianco

PR #1654, opened 2026-06-11
Support AddBiomechanics window in GUI and handle downloads in OpenSim format

aymanhab:
This PR adds an entry point to AddBiomechanics website to integrate into the OpenSim Application


aymanhab:
Build was successful on windows and included AddBiomechanics entry point, other artifacts were not generated so forcing a rebuild.

aymanhab:
The artifacts in this PR are ready for internal testing by actual addbiomechanics internal users. You should see addbiomechanics menu item under windows, from there it should be self explanatory. @mattpetrucci @AlbertoCasasOrtiz @nickbianco. Feel free to pass on to other interested parties. Thank you

Issue #1632, opened 2026-05-06
Support for Scholz2015GeometryPath in GUI "Epic"


aymanhab:
As of version 4.6, models with Scholz2015GeometryPath open in the GUI but do not visualize. This top issue will bring about the level of support we plan for version 4.6.1. In increasing levels of functionality/work:
1. Visualize path life and in animation, including color.
2. Path editing by changing properties of path "components"
3. Facilities/Editor to add/remove and/or move/drag components around.
4. Facilities to troubleshoot wrapping behavior.

aymanhab:
@nickbianco Starting point for the discussion regarding support for the new muscle wrapping in GUI

nickbianco:
Related API issue: https://github.com/opensim-org/opensim-core/issues/4279

aymanhab:
General outline, constraints to consider for visualization in GUI:
1. Scene graphs changes (add/remove) are expensive not practical so adding/removing points on the fly is not realistic.
2. Animations are time series sequences of transformations so all points should be visible all the time.

Here's how path animation works in GUI as of now
On model load: We find the geometryPath objects, foreach:
  - create "threejs:bones=anchor-points" per each potential (path)point.
  - create a cylinder mesh with the above "threejs:bones"

On every frame:
  - points anchored to bodies move for free 
  - bone positions need to be updated in one of two ways:
    - if wrapping is engaged, calculate new position
    - if not engaged, position points along the line segment they live on making them practically invisible (corner cases handling)

Ideally we use the same interfaces in AbstractGeometryPath regardless of which geometryPath is used but there're differences to consolidate, some unexposed functionality:
1. Old GeometryPath exposes all wrapObjects and where they fit on path, current sequence of elements Property is private.
2. Old GeometryPath has PathPoints property, now we get PathElements that we dereference or if we get PathPoints we lose their relation to obstracles.
3. Old GeometryPath was owned by muscle, is that true for Scholz2015GeometryPath?
4. We could deduce if a wrapObject is engaged or not, not clear for Scholz2015GeometryPath how to detect.

@nickbianco

nickbianco:
> bone positions need to be updated in one of two ways:

Do you mean points on a wrapping surface?

[...]

PR #1661, opened 2026-07-20
Opensim 46

aymanhab:
Fixes issue #0
### Brief summary of changes
merge opensim_46 into main

### CHANGELOG.md (choose one)

- no need to update because included

PR #1662, opened 2026-07-21
Visualize scholzpath

aymanhab:
Fixes issue #1632 

### Brief summary of changes
Handle visualization of ScholzGeometryPath instances, first draft that doesn't handle actual wrapping pending changes in opensim-core.

### Testing I've completed
Loaded model with Path and it visualizes properly.

### CHANGELOG.md (choose one)

- no need to update because WIP


aymanhab:
Once main branch is updated in PR #1661 there should be a single file change to review.

aymanhab:
For your review @nickbianco per your earlier request

nickbianco:
@aymanhab perhaps you could provide a high-level overview of the changes in the PR? It would be helpful to review since I'm not as familiar with the inner workings of the GUI.

aymanhab:
The one key class involved in this change is ModelVisualizationJson, it's used as follows:
- On model opening a Json description of the model is created and passed to the viewer. This is done by processing components one at a time. AbstractGeometryPath subclasses get special treatment since they don't map to visualization 1:1 all the time, check the method createJsonForScholzPath content and the call site for details.  We keep local maps so that we can update transforms as state changes.
- On every refresh/rendering of a state we call the method updateScholzPath with the new state to collect the new transforms and send them to the viewer in one json structure.
- When creating animations we call createAnimationJson() with a whole trajectory and generate an animation track for every moving object with the associated times and transforms.

There's significant book keeping to support selection/editing and the complicated old wrapping scenarios. 

The json object structures are native to three.js they are typically scene-graph nodes with associated geometry, material and transform/scale and a unique UUID. We try to reuse the geometry when possible for efficiency, reuse material so that color propagation is fast. 

Not sure how much of the above is useful for your review but please let me know if you have questions or suggestions to make future contributions/reviews more manageable.

In this PR there were a few places where I wanted to reuse the same book-keeping maps so I switched from GeometryPath to AbstractGeometryPath with the caveat that if we introduce a new subclass it may not visualize.

Gui/opensim/modeling/nbproject/private/private.xml
```diff
@@ -3,6 +3,18 @@
...
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/modeling/src/org/opensim/modeling/AbstractGeometryPath.java</file>
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/modeling/src/org/opensim/modeling/ExpressionBasedFunction.java</file>
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/modeling/src/org/opensim/modeling/OutputSpatialVec.java</file>
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/modeling/src/org/opensim/modeling/Logger.java</file>
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/modeling/src/org/opensim/modeling/Model.java</file>
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/modeling/src/org/opensim/modeling/C3DFileAdapter.java</file>
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/modeling/src/org/opensim/modeling/SWIGTYPE_p_std__functionT_void_fOpenSim__Component_const_p_SimTK__State_const_R_std__string_const_R_SimTK__VecT_2_SimTK__VecT_3_double_1_t_1_t_RF_t.java</file>
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/modeling/src/org/opensim/modeling/Component.java</file>
```
nickbianco, line 17: Is this debugging code? Why are these paths hard-coded?

Gui/opensim/tracking/nbproject/private/private.xml
```diff
@@ -17,6 +17,10 @@
         </file>
     </editor-bookmarks>
     <open-files xmlns="http://www.netbeans.org/ns/projectui-open-files/2">
-        <group/>
+        <group>
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/tracking/src/org/opensim/tracking/tools/InverseDynamicsToolAction.java</file>
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/tracking/src/org/opensim/tracking/InverseDynamicsToolModel.java</file>
+            <file>file:/D:/dev/src/opensim-gui/Gui/opensim/tracking/src/org/opensim/tracking/BaseToolPanel.java</file>
```
nickbianco, line 23: Same comment as above.

Gui/opensim/modeling/src/org/opensim/modeling/ScaleTool.java
```diff
@@ -78,20 +78,264 @@ public String getConcreteClassName() {
     return opensimActuatorsAnalysesToolsJNI.ScaleTool_getConcreteClassName(swigCPtr, this);
   }
 
-  public ScaleTool() {
-    this(opensimActuatorsAnalysesToolsJNI.new_ScaleTool__SWIG_0(), true);
+  public void copyProperty_mass(ScaleTool source) {
```
nickbianco, line 81: I'm assuming this is just ported bindings code and not explicitly related to the Scholz path changes?

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -156,7 +158,7 @@ public void setShowCom(boolean showCom) {
     private PathColorMap currentPathColorMap;
     // Preferences
     private double prefMuscleDisplayRadius=5;
-    private int NUM_PATHPOINTS_PER_WRAP_OBJECT=4;
+    private int NUM_PATHPOINTS_PER_WRAP_OBJECT=8;
```
nickbianco, line 161: Will increasing the number of points per object have an effect on the performance of visualizing a `GeometryPath`?

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -1846,9 +1920,94 @@ else if (secondIndex == -1){ // Conditional Path point that's inactive
...
+        UUID pathpt_mat_uuid = createPathPointMaterial(path);
+        // Cycle through pathElements and create reps for points, #NUM_WRAP_POINTS for each obstacle
+        // for now these are notr exposed and we don't know how to get them so will just assume no wrapping
+        // to explore downstream issues
+        JSONArray pathpoint_jsonArr = new JSONArray();
+        // Keep track of First path point so that when changing color we can propagate to all path points, since they share material
+        // Also so that when adding pathpoints due to wrapping, we can get proper/shared material
+        mapGeometryPathToPathPointMaterialUUID.put(path, pathpt_mat_uuid);
+        // Fix issue #2569 where moving pathpoint is not handled properly if first in path
+        int numObstacles  = path.getNumPathElements() - path.getNumPathPoints();
```
nickbianco, line 1934: [Scholz2015GeometrPath::getNumObstacles()](https://github.com/opensim-org/opensim-core/blob/cd5296d2a3e8d59ccf39079762f505cd563ffb23/OpenSim/Simulation/Model/Scholz2015GeometryPath.h#L427) is also available.

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -1846,9 +1920,94 @@ else if (secondIndex == -1){ // Conditional Path point that's inactive
...
+        UUID mat_uuid = createPathMaterial(path);
+        UUID pathpt_mat_uuid = createPathPointMaterial(path);
+        // Cycle through pathElements and create reps for points, #NUM_WRAP_POINTS for each obstacle
+        // for now these are notr exposed and we don't know how to get them so will just assume no wrapping
+        // to explore downstream issues
+        JSONArray pathpoint_jsonArr = new JSONArray();
+        // Keep track of First path point so that when changing color we can propagate to all path points, since they share material
+        // Also so that when adding pathpoints due to wrapping, we can get proper/shared material
+        mapGeometryPathToPathPointMaterialUUID.put(path, pathpt_mat_uuid);
+        // Fix issue #2569 where moving pathpoint is not handled properly if first in path
```
nickbianco: I don't think this comment is related to the code here?

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -1846,9 +1920,94 @@ else if (secondIndex == -1){ // Conditional Path point that's inactive
...
+            UUID scholzPathPointUUID = retrieveUuidFromJson(scholzPathPointJson);
+            pathpoint_jsonArr.add(scholzPathPointUUID.toString());
+            gndChildren.add(scholzPathPointJson);
+            //System.out.println("Adding a point at glround frame:");
+        }
+        pathsWithWrapping.put(path, pathpoint_jsonArr);
+        JSONObject pathGeomJson = new JSONObject();
+        UUID uuidForPathGeomGeometry = UUID.randomUUID();
+        pathGeomJson.put("uuid", uuidForPathGeomGeometry.toString());
+        pathGeomJson.put("type", "CylinderGeometry");
```
nickbianco, line 1970: I assume the cylinder geometry is used to draw lines?

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -418,6 +420,37 @@ private void addParentUuid(AbstractPathPoint firstPoint, JSONObject pointJson) {
 
     private static UUID retrieveUuidFromJson(JSONObject pointJson) {
         return UUID.fromString((String) pointJson.get("uuid"));
+    }
+
+    private JSONObject createScholzPathPointObjectJson(UUID pathpt_mat_uuid, DecorativeGeometry at) {
```
nickbianco, line 425: There are several new methods that are specific to the introduction of `Scholz2015GeometryPath`. Is this typical when adding new functionality to the GUI? I understanding needing some temporary shim code here and there (e.g., protecting against NaN path point values), but I would have expected a bit more generic treatment of paths, especially given the new decorative path point API in AbstractGeometryPath.

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
```diff
@@ -418,6 +420,37 @@ private void addParentUuid(AbstractPathPoint firstPoint, JSONObject pointJson) {
...
+        UUID objID = UUID.randomUUID();
+        ppt_object.put("uuid", objID.toString());
+        ppt_object.put("type", "Mesh");
+        ppt_object.put("geometry", geometryUuid.toString());
+        ppt_object.put("material", materialUuid.toString());
+        ppt_object.put("matrix", JSONUtilities.createMatrixFromTransform(xform, new Vec3(1.), 1.0));
+        return ppt_object;
+    }
+    // if scholzPath has non-engaged segment it returns NaNs, fix them
+    private void fixNanPathpointPositions(ArrayDecorativeGeometry adg) {
```
nickbianco, line 439: I've opened a Simbody issue to track this: https://github.com/simbody/simbody/issues/868

Gui/opensim/view/src/org/opensim/view/nodes/PropertyEditorAdaptor.java
```diff
@@ -122,6 +122,7 @@ public void handlePropertyChangeCommon() {
                 ContactGeometry.safeDownCast(obj)!=null){
             Component mc = Component.safeDownCast(obj);
             ViewDB.getInstance().updateComponentDisplay(model, mc, prop);
+            ViewDB.setApplyAppearanceChange(true);
```
nickbianco, line 125: Why was this line added?

nickbianco (commented):
I've left several comments below, none of them blocking since I don't yet have a full understanding of the inner workings of the GUI. However, some things like the hard-coded file paths and the ScaleTool API updates seem unrelated to adding support for Scholz2015GeometryPath.

I would also be concerned about the proliferation of code specifically tied to Scholz2015GeometryPath. Obviously it's not totally unavoidable, but given that both GeometryPath and Scholz2015GeometryPath essentially draw sequences of points and lines I would expect a bit more unified structure. Leveraging the new decorative path point API in AbstractGeometryPath could help.

aymanhab:
Thanks for the review @nickbianco 👍 
First I'll address the drive by changes:

- Hard coded paths in private.xml files are ignored by the platform. They were there before github and were always ignored, will add to gitignore in a separate PR. Thanks for reporting.

- Swig generated files are actually opensim-core changes funneled thru swig so that we're including the latest AbstractGeomertyPath, DisplayHints and generatedecoration related class changes etc. You can ignore unless something stands out. Typically I do these upgrades in separate PRs.

- Other drive-by one liners could be moved to separate PRs if needed.

Now the meat of the change is still in the one class ModelVisualizationJson. You'll see that very few OpenSim classes have special treatment other than GeometryPath and it's associated pathpoint subclasses which account for most of the complexity. Extra book keeping is due to the fact that the Gui has to wire a muscle/ligament to path and vice versa, which is not a visualizer functionality.  Additional complexity reasons are:

1. The old wrapping classes have been untouched in opensim-core since version 1.0 except for minor optimizations here or there on the basis that they'll be replaced at some point and that there're a ton of corner cases baked into them since SIMM days that we didn't want to break for backward compatibility.
2. There are major differences between the old and new wrapping algorithms for example: conditional path points that can engage, disengage. Path can contact with wrap object multiple times within one or in different segments. Include Quadrants as well for fun and try to represent the result using fixed number of points (so that scene graph is not structurally changed on the fly) and you'll get a feel for what the visualization code needed to handle. the new decorative pathpoint doesn't change this picture except for marking some spheres as decorative pathpoints, unless we mess around with the point generation process this has little effect downstream.
3. The changes to support the new wrapping are relatively small thanks to the new cleaner design. We have clear ordering and fixed number of segments that either engage or disengage. Some shared data structures for bookkeeping were updated to use AbstractGeometryPath so they could be reused. I can look into more unification that doesn't require major changes to the old path visualization that we hope to phase out.
4.  Indeed the number of samples increasing was to account for finer discretization, in practice the difference is minimal for the models I tried. Some test models of the new wrapping had longer paths so the discretization was more sparse but this is up for discussion.


Great to have this discussion here and to have you familiarized with the visualization pipeline used by the GUI.

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
aymanhab: I totally understand, unfortunately the issue is more involved as I described below

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
aymanhab: Indeed, and in general we can use any mesh shape since this is skinning that can be applied to 3d muscles if we have meshes for them for example to inflate or change shape on activation change.

Gui/opensim/view/src/org/opensim/threejs/ModelVisualizationJson.java
aymanhab: Not significantly in the big picture, I touched upon that in my comment below

nickbianco:
Thanks for the extra clarification @aymanhab! Good to have a better handle on the GUI machinery. 

Again, none of my comments are blocking, and it seems like there's sufficient reasoning for the changes here. Glad that the new wrapping class keeps the implementation on the GUI side relatively minimal compared to `GeometryPath`.

Issue #1568, opened 2025-11-07
Workflow dispatch from opensim-core fails due to credentials issue


nickbianco:
The error message from GitHub Actions:

```
Run codex-/return-dispatch@v1
  with:
    token: ***
    ref: refs/heads/main
    repo: opensim-gui
    owner: opensim-org
    workflow: build-on-core-pr.yml
    workflow_timeout_seconds: 300
  env:
    GH_TOKEN: ***
Fetching Workflow ID for build-on-core-pr.yml...
Error: getWorkflowId: An unexpected error has occurred: Bad credentials - https://docs.github.com/rest
Error: Failed: Bad credentials - https://docs.github.com/rest
Warning: Does the token have the correct permissions?
Error: Bad credentials - https://docs.github.com/rest
```

PR #1665, opened 2026-07-29
decompose workfow for reuse

aymanhab:
Fixes issue #1568 

### Brief summary of changes
Decompose main workflow file and reuse in workflow invoked from opensim-core 

### Testing I've completed
Ran ci

### CHANGELOG.md (choose one)


aymanhab:
First step in the direction of modernizing ci, next up using presets and handle osx

PR #1677, opened 2026-09-10
Update bindings using swig 4.4.1 now

aymanhab:
Fixes issue #0

### CHANGELOG.md (choose one)

- updated...


aymanhab:
There's no user code here just upgrading swig and opensim-core code base.

